PR #1, opened 2018-10-02
Add packet fragmentation

TimonPost:
This is an implementation of fragmentation ([issue](https://github.com/LucioFranco/amethyst-protocol/issues/5)). Looks like it is working, but need to test it more still. So work in progress...

Note that I have looked at this [libary](https://github.com/jaynus/netcode.io/tree/master/rust) which had some basic implementation of fragmentation. It is from the author of gafferongames blogs: 

**Things I changed:**

1. There are two headers:

- PacketHeader
This will contain all the header stuff we first had in  RawPacket. This will be used for every packet.

- Fragment Header
Fragment header the information of the fragments a packet is divided into. The first fragment packet will have two headers. The standard PacketHeader and the Fragment header the following fragments only have the fragment header. 

2. Also, I have changed to using `byteorder` so that I can control more better how a header gets serialized an deserialized.
3. I have a FragmentBuffer which will be used to store fragments temporarily
4. Added a config ty[e for network settings.

**Still, need to do:**
1. Tests
2. Config file needs to be given by the user (for now the default is used).
3. I want to make the preproccess_packet function smaller it is pritty big now.


src/net/udp.rs
```diff
@@ -1,48 +1,61 @@
...
+    config: NetworkConfig,
+    packet_processor: PacketProcessor
 }
 
 impl UdpSocket {
     pub fn bind<A: ToSocketAddrs>(addr: A) -> io::Result<Self> {
         let socket = net::UdpSocket::bind(addr)?;
         let state = SocketState::new();
+        // TODO: add functionality to get config from file.
+        let config = NetworkConfig::default();
```
LucioFranco: I think we should just pass this in, let the user decide how to create the config.

src/net/udp.rs
```diff
@@ -1,48 +1,61 @@
...
     }
 
-    pub fn send(&mut self, packet: Packet) -> Result<io::Result<usize>> {
-        let (addr, payload) = self.state.pre_process_packet(packet)?;
-        Ok(self.socket.send_to(&payload, addr))
+    /// Sends data on the socket to the given address. On success, returns the number of bytes written.
+    pub fn send(&mut self, mut packet: Packet) -> Result<usize> {
+        let (addr, mut packet_data) = self.state.pre_process_packet(packet, &self.config)?;
+
+        let mut bytes_send = 0;
```
LucioFranco: `bytes_sent`

src/packet/header/fragment.rs
```diff
@@ -0,0 +1,128 @@
...
+        wtr.write_u8(self.num_fragments)?;
+
+        if self.id == 0 {
+            match self.packet_header
+            {
+                Some(header) => {
+                    wtr.write(&header.parse()?)?;
+                },
+                None => {
+                    return Err(Error::new(ErrorKind::Other, "Invalid fragment header"));
```
LucioFranco: Should have an error type for this

src/packet/header/packet.rs
```diff
@@ -0,0 +1,85 @@
...
+            seq: seq_num,
+            ack_seq: last_seq,
+            ack_field: bit_field,
+        }
+    }
+
+    /// Get the size of this header.
+    pub fn size(&self) -> u8
+    {
+        return PACKET_HEADER_SIZE;
```
LucioFranco: You dont need the return keyword here 😄

src/packet/packet_processor.rs
```diff
@@ -0,0 +1,220 @@
...
+        let mut received_bytes = Ok(None);
+
+        if prefix_byte & 1 == 0 {
+            received_bytes = self.handle_normal_packet(&mut cursor, &addr, socket_state);
+        } else {
+            received_bytes = self.handle_fragment(&mut cursor);
+        }
+
+        return match received_bytes {
+            Ok(Some(payload)) => return Ok(Some(Packet::new(addr, payload))),
```
LucioFranco: drop the return here

src/packet/packet_processor.rs
```diff
@@ -0,0 +1,220 @@
...
+            Ok(None) => return Ok(None),
```
LucioFranco: Same as above

src/net/socket_state.rs
```diff
@@ -40,30 +42,63 @@ impl SocketState {
...
-            l.their_acks.last_seq,
-            l.their_acks.field,
-        );
-        // increase sequence number
-        l.seq_num = l.seq_num.wrapping_add(1);
-        let buffer = serialize(&raw_packet)?;
-        Ok((packet.addr, buffer))
+        let connection_seq = lock.seq_num;
+        // queue new packet
+        lock.waiting_packets.enqueue(connection_seq, packet.clone());
```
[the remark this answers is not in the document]
TimonPost: For now, it does not really matter much though because the packet is not changed after it is enqueued waiting for acknowledgment. The packet is content will be ready before it is enqueued. What we can do to avoid change is to make the fields of Packet nonpublic and only readable. So once a Packet is defined you can change it anymore.

src/net/socket_state.rs
```diff
@@ -40,30 +42,63 @@ impl SocketState {
...
+
+            if num_fragments > config.max_fragments {
+                return Err(NetworkError::ExceededMaxFragments.into());
+            }
+
+            for fragment_id in 0..num_fragments {
+                let fragment = FragmentHeader::new(fragment_id, num_fragments, packet_header.clone());
+
+                // get start end pos in buffer
+                let start_fragment_pos = fragment_id as u16 * config.fragment_size; /* upcast is safe */
```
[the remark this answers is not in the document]
TimonPost: We do store it the fragment_size as u16. I think you mean fragment_id? But that is an u8 so the upcast is safe here. To multiply fragment_id with fragment size they need to be of the same type. And by casting the fragment_id to u16 we can multiply those two numbers.

src/packet/header/fragment.rs
TimonPost: There is one problem of doing that you have to see the context of the whole function. 

Every 'write' in that function returns an `io::Result`. So we could paste this line `.map_err(|| NetworkError::InvalidPacketHeader.into()?);` behind each write function or we could simply return `io::Result` and let the caller of that function maps the error to `.map_err(|| NetworkError::InvalidPacketHeader.into()?);` 

I think that is better otherwise this function will look horrible.

src/net/socket_state.rs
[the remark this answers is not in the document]
TimonPost: I have changed all fields of the packet to private and made getters for them so they can not be changed once declared.

src/error.rs
```diff
@@ -16,5 +16,21 @@ pub enum NetworkError {
...
     #[fail(display = "The lock for a specific TCP client was poisoned")]
-    TcpClientLockFailed
+    TcpClientLockFailed,
+    #[fail(display = "The fragment of an packet is invalid")]
+    InvalidFragmentHeader,
+    #[fail(display = "The parsing of the header went wrong")]
+    HeaderParsingFailed,
+    #[fail(display = "Something went wrong when sending")]
+    SendFailed,
+    #[fail(display = "Something went wrong when receiving")]
```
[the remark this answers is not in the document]
TimonPost: we will do this later

src/error.rs
```diff
@@ -16,5 +16,21 @@ pub enum NetworkError {
...
+    ReceiveFailed,
+    #[fail(display = "The packet size was bigger than the max allowed size.")]
+    ExceededMaxPacketSize,
+    #[fail(display = "The total of fragments the packet was divided into is bigger than the allowed fragments.")]
+    ExceededMaxFragments,
+    #[fail(display = "The packet header is invalid.")]
+    InvalidPacketHeader,
+    #[fail(display = "Just an error where we don't care about the type.")]
```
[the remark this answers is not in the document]
TimonPost: Yea can be removed had it for test forgot to remove it.

Cargo.toml
```diff
@@ -10,6 +10,7 @@ license = "MIT/Apache-2.0"
 serde = "1.0"
 serde_derive = "1.0"
 bincode = "1.0"
+byteorder = "*"
```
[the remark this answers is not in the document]
TimonPost: no, no reason going to set it to an version good point!

src/net/socket_state.rs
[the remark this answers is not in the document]
LucioFranco: Okay, thats fine, I was just thinking that this would be a good use case for a ref counter.

src/lib.rs
```diff
@@ -18,4 +18,39 @@ pub mod error;
...
+///
+/// Calculate modules:
+/// - number 1 % number 2 = 928
+/// - this is bigger than 0 so remainder = 1
+///
+/// Calculate how many times the `number 1` fits in `number 2`:
+/// - number 1 / number 2 = 3,90625 (this will be rounded to 3)
+/// - add remainder from above to 3 = 4.
+///
+/// An you will that for all numbers you push into this method it will return an nice result.
```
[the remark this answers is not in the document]
TimonPost: Gotcha must be copy paste work :)

src/packet/header/fragment.rs
LucioFranco: In reality we should have a Error::from implemented for `io::Result`

Cargo.toml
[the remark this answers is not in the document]
LucioFranco: Agreed.

src/net/socket_state.rs
```diff
@@ -40,30 +42,63 @@ impl SocketState {
...
+            error!("Packet too large: Attempting to send {}, max={}", packet.payload().len(), config.max_packet_size);
+            return Err(NetworkError::ExceededMaxPacketSize.into());
         }
 
-        let raw_packet: RawPacket;
-        // initialize packet data, seq, acked_seq etc.
-        let mut l = connection
+        let connection = self.create_connection_if_not_exists(&packet.addr)?;
+
+        let mut lock = connection
```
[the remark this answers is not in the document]
TimonPost: fixed!

src/net/constants.rs
```diff
@@ -0,0 +1,3 @@
+pub const DEFAULT_MTU: usize = 1452;
```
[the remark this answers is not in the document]
TimonPost: Jup, further investigation needed. I'll keep it like this for now. If you wondered why 1452 (gaffer recommended size): 

/// Derived from ethernet_mtu - ipv6_header_size - udp_header_size - packet header size
///       1452 = 1500         - 40               - 8               - 8

src/packet/fragment_buffer.rs
```diff
@@ -0,0 +1,109 @@
...
+            entry_sequences,
+        }
+    }
+
+    /// Get mutable entry from collection by sequence number.
+    pub fn get_mut(&mut self, sequence: u16) -> Option<&mut T> {
+        let index = self.index(sequence);
+
+        if self.entry_sequences[index] != sequence {
+            return None;
```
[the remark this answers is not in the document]
TimonPost: We need to return here other wise I'll get the error expected `()` found Option

src/error.rs
```diff
@@ -1,9 +1,17 @@
 use failure;
 use std::result;
+use std::io::{self, ErrorKind};
 
 pub type Error = failure::Error;
 pub type Result<T> = result::Result<T, Error>;
 
+//impl From<io::Error> for Error
```
fhaynes: Can we not leave commented code lying about? I know we're going to redo error-handling, but I think its good habit in general to not leave commented out code in the code, and instead put it in an issue

src/net/network_config.rs
```diff
@@ -1,21 +1,39 @@
...
 #[derive(Clone)]
 pub struct NetworkConfig
 {
+    /// This is the maximal size an packet can get with all its fragments.
+    ///
+    /// Recommended value: 16384
     pub max_packet_size: usize,
+    /// These are the maximal fragments a packet could be divided into.
+    ///
+    /// Why can't I have more than 256 (u8)?
```
torkleyy: super minor nit: 255

src/net/network_config.rs
```diff
@@ -1,21 +1,39 @@
...
+    /// This is because you don't want to send more then 256 fragments over UDP, with high amounts of fragments the change for an invalid packet is very high.
```
torkleyy: "change" -> "chance"

Dumb question from somebody who doesn't know much about networking: If one packet failed, couldn't we just resend it? I think it would be ideal if the size of the packet can be as large as necessary, so the user does not have to split very big messages manually.

TimonPost: One time stuff? What we want to have like TCP: ordered +  able to see dropped packets (not automatically resent them) + the fragmentation (this PR, and later to be able to send each fragment individually if dropped).

fhaynes: @torkleyy You can also think of it this way: TCP is built on top of UDP. We're doing the same thing, but only implementing the few features of TCP that are useful to a network game.

Regarding packet size, when sending data over the Internet, in general you want to keep the packet size at less than 1500 bytes. The reason for this is historical. Thus, we have to implement our own fragmentation, since plain UDP does not have it.

torkleyy: @TimonPost I meant things like assets. I'm just reading all the documents and @fhaynes answers my questions on Discord ;) I understood this now, you can see this as resolved except for "change" which should be "chance".

TimonPost: I'll have already changed that thanks for looking into the PR

src/lib.rs
```diff
@@ -3,19 +3,54 @@
...
+/// This functions checks how many times a number fits into another number and will round up.
+///
+/// For example we have two numbers:
+/// - number 1 = 4000;
+/// - number 2 = 1024;
+/// If you do it the easy way the answer will be 4000/1024 = 3.90625.
+/// But since we care about how how many whole times the number fits in we need the result 4.
+///
+/// Note that when rust is rounding it is always rounding to zero (3.456 as u32 = 3)
+/// 1. calculate with modules if `number 1` fits exactly in the `number 2`.
```
torkleyy: not sure if you meant "modulo"?

src/lib.rs
```diff
@@ -3,19 +3,54 @@
...
+///
+/// The above described method will figure out for all number how many times it fits into another number rounded up.
+///
+/// So an example of dividing an packet of bytes we get these fragments:
+///
+/// So for 4000 bytes whe need 4 fragments
+/// [fragment: 1024] [fragment: 1024] [fragment: 1024] [fragment: 928]
+fn total_fragments_needed(payload_length: u16, fragment_size: u16) -> u16
+{
+    let remainder = if payload_length % fragment_size > 0 { 1 } else { 0 };
```
torkleyy: How about `max((payload_length + fragment_size - 1) / fragment_size, 1)`?

src/packet/header/fragment.rs
```diff
@@ -0,0 +1,151 @@
...
+}
+
+impl FragmentHeader {
+    /// Create new fragment with the given packet header
+    pub fn new(id: u8, num_fragments: u8, packet_header: PacketHeader) -> Self {
+        FragmentHeader { id, num_fragments, packet_header: Some(packet_header), sequence: packet_header.seq }
+    }
+
+    /// Get the size of this header.
+    pub fn size(&self) -> u8
```
torkleyy: These methods should probably go into a trait.

TimonPost: Good point going to do that.

src/lib.rs
TimonPost: I come back at this need to take a look if this works.

src/error.rs
```diff
@@ -1,9 +1,13 @@
 use failure;
 use std::result;
+use std::io::{self, ErrorKind};
 
 pub type Error = failure::Error;
 pub type Result<T> = result::Result<T, Error>;
 
+
```
LucioFranco: Remove these new lines please

src/packet/fragment_buffer.rs
```diff
@@ -0,0 +1,109 @@
+use std::io::Result;
+use std::clone::Clone;
+
+/// Collection for storing data fof any kind.
```
LucioFranco: `fof` should be `of`

src/packet/fragment_buffer.rs
```diff
@@ -0,0 +1,109 @@
...
+
+        self.entries[index] = data;
+        self.entry_sequences[index] = sequence;
+
+        Ok(&mut self.entries[index])
+    }
+
+    /// Remove entry from collection.
+    pub fn remove(&mut self, sequence: u16) {
+        // TODO: validity check
```
LucioFranco: Could you explain more what you mean by this `TODO`.

LucioFranco:
@fhaynes btw
````
error: expected `;`, found `use`
 --> src/net/udp.rs:5:1
  |
4 | use super::{constants, Packet, RawPacket}
  |                                          - expected `;`
5 | use socket_state::SocketState; 
  | ^^^ unexpected token
```

fhaynes:
> @fhaynes btw
> 
> ```
> error: expected `;`, found `use`
>  --> src/net/udp.rs:5:1
>   |
> 4 | use super::{constants, Packet, RawPacket}
>   |                                          - expected `;`
> 5 | use socket_state::SocketState; 
>   | ^^^ unexpected token
> ```
> ```

Fixed, thanks.

TimonPost:
Oke fixed the merge, ran all the tests with success, removed some unused imports and fixed all your comments. So going to merge now since you both have approved.

PR #5, opened 2018-10-06
Add server_client example

torkleyy:
This is a stupid example, yes. I'm probably doing something wrong, because I noticed two issues:

* messages take very long (~ 1 second) until they're shown. probably some blocking going on somewhere?
* when there is no server and I send a message with the client, I don't get any error message (nothing happens)

### DIFF
--- a/examples/server_client.rs
+++ b/examples/server_client.rs
@@ -1,0 +1,99 @@
+//! Note that the terms "client" and "server" here are purely what we logically associate with them.
+//! Technically, they both work the same.
+//! Note that in practice you don't want to implement a chat client using UDP.
+
+extern crate laminar;
+
+use std::io::stdin;
+
+use laminar::{
+    error::Result,
+    NetworkConfig,
+    Packet,
+    UdpSocket,
+};
+
+const SERVER: &str = "localhost:12351";
+
+fn server() -> Result<()> {
+    let mut socket = UdpSocket::bind(SERVER, NetworkConfig::default())?;
+
+    println!("Listening for connections to {}", SERVER);
+
+    loop {
+        match socket.recv()? {
+            Some(packet) => {
+                let msg = packet.payload();
+
+                if msg == b"Bye!" {
+                    break;
+                }
+
+                let msg = String::from_utf8_lossy(msg);
+                let ip = packet.addr().ip();
+
+                println!("Received {:?} from {:?}", msg, ip);
+
+                socket.send(Packet::new(packet.addr(), "Copy that!".as_bytes().to_vec()))?;
+            }
+            None => {}
+        }
+    }
+
+    Ok(())
+}
+
+fn client() -> Result<()> {
+    let mut socket = UdpSocket::bind("localhost:12352", NetworkConfig::default())?;
+
+    let server = SERVER.parse()?;
+
+    println!("Type a message and press Enter to send. Send `Bye!` to quit.");
+
+    let stdin = stdin();
+    let mut s_buffer = String::new();
+
+    loop {
+        s_buffer.clear();
+        stdin.read_line(&mut s_buffer)?;
+        let line = s_buffer.replace(|x| x == '\n' || x == '\r', "");
+
+        socket.send(Packet::new(server, line.clone().into_bytes()))?;
+
+        if line == "Bye!" {
+            break;
+        }
+
+        let back = socket.recv()?;
+
+        match back {
+            Some(packet) => {
+                if packet.addr() == server {
+                    println!("Server sent: {}", String::from_utf8_lossy(packet.payload()));
+                } else {
+                    println!("Unknown sender.");
+                }
+            }
+            None => println!("Silence.."),
+        }
+    }
+
+    Ok(())
+}
+
+fn main() -> Result<()> {
+    let stdin = stdin();
+
+    println!("Please type in `server` or `client`.");
+
+    let mut s = String::new();
+    stdin.read_line(&mut s)?;
+
+    if s.starts_with("s") {
+        println!("Starting server..");
+        server()
+    } else {
+        println!("Starting client..");
+        client()
+    }
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,10 +1,10 @@
 [package]
 name = "laminar"
-version = "0.0.0"
+version = "0.1.0"
 authors = [
-  "Lucio Franco <luciofranco14@gmail.com>",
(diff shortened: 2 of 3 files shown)


torkleyy:
**Not meant to be merged**

TimonPost:
I did some test to monitor how long it took to send a message and it took 1 second. So there is some bug somewhere :)

test results: 

```
= Message took 72.982µs to send =
Moving to lat: 10.555, long: 10.55454, alt: 1.3
=Message took 990.276878ms to send =
Moving to lat: 5.4545, long: 3.344, alt: 1.33
= Message took 990.114451ms to send =
Received text: "Some information"
```

After debugging I found that `create_connection_if_not_exists` method takes 0.5 - 0.9 seconds to execute.

And found out that acquiring the lock on `connections` is taking to long in `create_connection_if_not_exists` method. 

```
  let mut lock = self
            .connections
            .write()
            .map_err(|_| NetworkError::AddConnectionToManagerFailed)?;
```

examples/server_client.rs
```diff
@@ -0,0 +1,100 @@
...
+        }
+    }
+
+    Ok(())
+}
+
+fn client() -> Result<()> {
+    let mut socket = UdpSocket::bind("localhost:12352")?;
+
+    let server = addr(SERVER)?;
```
TimonPost: Use let server: `SocketAddr = SERVER.parse().unwrap()` instead

examples/server_client.rs
```diff
@@ -0,0 +1,100 @@
...
+
+use laminar::{
+    error::Result,
+    Packet,
+    UdpSocket,
+};
+
+const SERVER: &str = "localhost:12351";
+
+fn addr(s: &str) -> Result<SocketAddr> {
```
TimonPost: Remove this and replace it with  `let addr: SocketAddr = string.parse().unwrap();` See comment below.

examples/server_client.rs
```diff
@@ -0,0 +1,100 @@
...
+    loop {
+        match socket.recv()? {
+            Some(packet) => {
+                let msg = packet.payload();
+
+                if msg == b"Bye!" {
+                    break;
+                }
+
+                let msg = String::from_utf8_lossy(msg);
```
TimonPost: Why not check here if message is `Bye!` with `msg`.

examples/server_client.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    let stdin = stdin();
+    let mut s = String::new();
+
+    loop {
+        s.clear();
+        stdin.read_line(&mut s)?;
+        let line = s.replace(|x| x == '\n' || x == '\r', "");
+
+        socket.send(Packet::new(server, line.clone().into_bytes()))??;
```
TimonPost: `??`?

examples/server_client.rs
```diff
@@ -0,0 +1,100 @@
...
+
+        if line == "Bye!" {
+            break;
+        }
+
+        let back = socket.recv()?;
```
TimonPost: what is back?

examples/server_client.rs
```diff
@@ -0,0 +1,100 @@
...
+
+fn client() -> Result<()> {
+    let mut socket = UdpSocket::bind("localhost:12352")?;
+
+    let server = addr(SERVER)?;
+
+    println!("Type a message and press Enter to send. Send `Bye!` to quit.");
+
+    let stdin = stdin();
+    let mut s = String::new();
```
TimonPost: Please some better name ;)

src/net/socket_state.rs
```diff
@@ -94,28 +92,19 @@ impl SocketState {
...
-                    if last_heard >= sleepy_time {
-                        // TODO: pass up client TimedOut event
-                        error!("Client has timed out: {:?}", key);
-                    }
-                }
-
-                thread::sleep(poll_interval)
-            }).unwrap();
+    /// Removes all the connections that haven't sent packets for at least `self.timeout`.
+    pub fn remove_timed_out(timeout: u64,
```
TimonPost: other name would be better since you don't remove timed out connections here.

src/net/socket_state.rs
```diff
@@ -160,11 +158,4 @@ mod test {
         let addr = format!("{}:{}", TEST_BAD_HOST_IP, TEST_PORT).to_socket_addrs();
         assert!(addr.is_err());
     }
-
-    #[test]
-    fn test_poll_for_invalid_clients() {
```
TimonPost: removing test because they fail :+1:. Actually don't think this test adds much value. So it's fine with me

torkleyy: This ond didn't do anything. There's another one that fails, need to check why. For this test I'm gonna add a better one.

examples/server_client.rs
```diff
@@ -0,0 +1,100 @@
...
+fn main() -> Result<()> {
+    let stdin = stdin();
+
+    println!("Please type in `server` or `client`.");
+
+    let mut s = String::new();
+    stdin.read_line(&mut s)?;
+
+    if s.starts_with("s") {
+        server()
```
LucioFranco: I would add a println saying the server is starting just for clairity

examples/server_client.rs
```diff
@@ -0,0 +1,100 @@
...
+    } else {
+        client()
```
LucioFranco: Same as above.

torkleyy: For the client there is one println already that it started, do you wanto a `Starting..`, too?

examples/server_client.rs
torkleyy: The message that we get _back_ ;)

examples/server_client.rs
torkleyy: Yes, those are result wrapped results, most likely for non-blocking stuff.

examples/server_client.rs
torkleyy: doesn't really matter I guess?

examples/server_client.rs
LucioFranco: No thats fine.

examples/server_client.rs
TimonPost: Yes but you convert to times bytes to string. Just a little thing.

torkleyy: Sorry, I don't understand. I just did it that way because checking for that byte sequence seemed more accurate than checking a lossy conversion that was intended only for printing. Of course it'll be the same this case, that's just why I did it if that makes any sense.

TimonPost: Oke lets leaf it like that.

PR #6, opened 2018-10-06
Added library examples

TimonPost:
Added examples. 

I noticed that the API was not cool to use so I have made some changes in that too. It now looks more like the `std::net::` API. 

**Things changed:**

1. name space change:
```
extern crate laminar;

use self::laminar::net::{UdpSocket, TcpClient, TcpServer, TcpSocketState, SocketAddr};
```
2. method name change:
So first the udp socket had an method called `set_nonblocking(bool)` this confuses me every time I use it. I have changed this to `set_blocking(bool)` witch looks to my opinion much cleaner.
3. changed `use super::Packet` and all usings to `Packet` with `packet::Packet`
3. Added examples
following file structure:
- examples 
    - mod.rs 
      File that contains a method that could be runned with `cargo test --example playground`. So now you can play directly around with the examples without needing to create a new project to test the library. 
    - tcp.rs
      This file will contain the basic examples for the TCP Api.
    - udp.rs
     This file will contain the basic examples for the UDP Api.
    - simple_udp.rs
     This file contain a simple UDP server client program.


LucioFranco:
@TimonPost there are some conflicts, I forgot to push the new `cargo.toml` so please fix that.

examples/tcp.rs
```diff
@@ -0,0 +1,7 @@
+//! This module provides examples for the TCP api.
+
+extern crate laminar;
+
+use self::laminar::net::{UdpSocket, TcpClient, TcpServer, TcpSocketState, SocketAddr};
+
+// TODO
```
LucioFranco: If we are gonna leave this TODO here, lets remove this file, and open an issue, instead.

examples/udp.rs
```diff
@@ -0,0 +1,100 @@
...
+pub fn receive_data_with_blocking()
+{
+    // setup an udp socket and bind it to the client address.
+    let mut udp_socket: UdpSocket = UdpSocket::bind(server_address()).unwrap();
+
+    // next we could specify if or socket should block the current thread when receiving data or not (default = true)
+    udp_socket.set_blocking(true);
+
+    // Next start receiving.
+    let result= udp_socket.recv();
```
LucioFranco: space before the `=`

examples/udp.rs
```diff
@@ -0,0 +1,100 @@
...
+{
+    // setup an udp socket and bind it to the client address.
+    let mut udp_socket: UdpSocket = UdpSocket::bind(client_address()).unwrap();
+
+    // next we could specify if or socket should block the current thread when receiving data or not (default = true)
+    udp_socket.set_blocking(false);
+
+    // setup a thread to do the receiving
+    // Next start receiving.
+    let result= udp_socket.recv();
```
LucioFranco: Same as above.

src/net/udp.rs
```diff
@@ -43,8 +43,8 @@ impl UdpSocket {
         Ok(self.socket.send_to(&payload, addr))
     }
 
-    pub fn set_nonblocking(&mut self, nonblocking: bool) -> io::Result<()> {
-        self.socket.set_nonblocking(nonblocking)
+    pub fn set_blocking(&mut self, blocking: bool) -> io::Result<()> {
```
LucioFranco: Why this change of api? I wanted to originally follow the same api as provided in the stdlib.

src/lib.rs
```diff
@@ -49,7 +49,4 @@ fn total_fragments_needed(payload_length: u16, fragment_size: u16) -> u16
 {
     let remainder = if payload_length % fragment_size > 0 { 1 } else { 0 };
      ((payload_length / fragment_size) + remainder)
-}
-
-pub use net::{NetworkConfig, UdpSocket};
-pub use packet::Packet;
```
torkleyy: Why are you removing those again? They're so common I thought they make sense here.

TimonPost: I did that to make the user API better. Like the standard lib has:

use laminar::net::{all socket stuff}
use laminar::packet::{packet stuff}

I don't like it when we re export some types and other types not. Also when importing those types I want to be consistent instead of using super:: Packet for example.

TimonPost:
I have one comment to add to why I did that @LucioFranco. 

following file structure:
- examples 
    - mod.rs 
      File that contains a method that could be ran with `cargo test --example playground`. So now you can play directly around with the examples without needing to create a new project to test the library. 
    - tcp.rs
      This file will contain the basic examples for the TCP Api.
    - udp.rs
     This file will contain the basic examples for the UDP Api.
    - simple_udp.rs
     This file contain a simple UDP server client program. 

This way we have actualy all the examples seperated. But why the mod file than? I did this because of two reasons:

1. Because you can play arround with and test  the examples in the `mod.rs` and run this with `cargo run --example playground`. 
2. Because you test here things like the user api of the crate things you made etc. in mod.rs. You than run it with `cargo run --example playground`. 

It is basically a playground for developing purpose and for testing etc.

If you want to see an example of how I did it with my own crate see [this](https://github.com/TimonPost/crossterm/tree/master/examples)
I

PR #9, opened 2018-10-06
7 sending takes too long

fhaynes:
This fixes the background timeout check thread from holding on to the read lock while it sleeps. It also adds comments to the public functions, and normalizes the errors to use our error module.

### DIFF
--- a/src/net/socket_state.rs
+++ b/src/net/socket_state.rs
@@ -22,22 +22,18 @@ const TIMEOUT_POLL_INTERVAL: u64 = 1;
 pub struct SocketState {
     timeout: ConnectionTimeout,
     connections: ConnectionMap,
-    timeout_check_thread: Option<thread::JoinHandle<()>>
+    timeout_check_thread: thread::JoinHandle<()>
 }
 
 impl SocketState {
     pub fn new() -> Result<SocketState> {
-        let mut socket_state = SocketState {
+        let connections: ConnectionMap = Arc::new(RwLock::new(HashMap::new()));
+        let thread_handle = SocketState::check_for_timeouts(connections.clone())?;
+        Ok(SocketState {
             connections: Arc::new(RwLock::new(HashMap::new())),
             timeout: TIMEOUT_DEFAULT,
-            timeout_check_thread: None
-        };
-        if let Ok(thread_handle) = socket_state.check_for_timeouts() {
-            socket_state.timeout_check_thread = Some(thread_handle);
-            Ok(socket_state)
-        } else {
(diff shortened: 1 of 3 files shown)


src/net/socket_state.rs
```diff
@@ -22,16 +22,22 @@ const TIMEOUT_POLL_INTERVAL: u64 = 1;
...
 }
 
 impl SocketState {
-    pub fn new() -> SocketState {
+    pub fn new() -> Result<SocketState> {
         let mut socket_state = SocketState {
             connections: Arc::new(RwLock::new(HashMap::new())),
             timeout: TIMEOUT_DEFAULT,
+            timeout_check_thread: None
         };
```
LucioFranco: New line here would be nice.

src/net/socket_state.rs
```diff
@@ -94,28 +100,40 @@ impl SocketState {
...
             .spawn(move || loop {
-                let connections = connections_lock.read().expect("Unable to aquire read lock");
-
-                for (key, value) in connections.iter() {
-                    let connection = value.read().expect("Unable to aquire read lock");
-                    let last_heard = connection.last_heard();
-                    if last_heard >= sleepy_time {
-                        // TODO: pass up client TimedOut event
-                        error!("Client has timed out: {:?}", key);
+                {
```
LucioFranco: Are you sure you need this extra scope here? The loop scope should cover the dropping of the lock.

src/net/socket_state.rs
```diff
@@ -94,28 +100,40 @@ impl SocketState {
...
+                    debug!("Checking for timeouts");
```
LucioFranco: I would switch this to trace.

src/net/socket_state.rs
```diff
@@ -22,16 +22,18 @@ const TIMEOUT_POLL_INTERVAL: u64 = 1;
...
     connections: ConnectionMap,
+    timeout_check_thread: thread::JoinHandle<()>
 }
 
 impl SocketState {
-    pub fn new() -> SocketState {
-        let mut socket_state = SocketState {
+    pub fn new() -> Result<SocketState> {
+        let connections: ConnectionMap = Arc::new(RwLock::new(HashMap::new()));
+        let thread_handle = SocketState::check_for_timeouts(connections.clone())?;
```
LucioFranco: new line after this please.

fhaynes: I'm just going to run cargo fmt. =P

LucioFranco: 👍

Issue #11, opened 2018-10-07
Decide how to bubble up events from laminar to the network crate


fhaynes:
# Idea 1
Pass an mpsc channel down, that carries an enum with message variants?

fhaynes:
So the events we're going to care about in laminar can be expressed, I think, like this:
```
enum LaminarEvent {
    ClientConnect{ip: String, port: String, payload: &[u8]},
    PacketReceived{payload: &[u8]}
    ClientTimeout{ip: String, port: String}
}
```
The payloads are slices into the payload section of the Packet, because the Network crate doesn't care about the TCP/UDP headers. More coming in the next comment...

fhaynes:
Then in the recv function of udp.rs, something like:
```
    pub fn recv(&mut self) -> Result<Option<Packet>> {
        let (len, _addr) = self.socket.recv_from(&mut self.recv_buffer)?;
        // If len is 0, then there was no data in the packet
        if len > 0 {
            // If there was data, deserialize it and hand it off for processing
            let raw_packet: RawPacket = deserialize(&self.recv_buffer[..len])?;
            let packet = self.state.process_received(_addr, &raw_packet)?;
            self.chan.send(LaminarEvent::PacketReceived{ip: &packet.ip, addr: &packet.addr, payload: &packet.payload});
            Ok(Some(packet))
        } else {
            Ok(None)
        }
    }
```

fhaynes:
The same pattern can be used in the timeout thread or anywhere else to generate events and send them to the receiver, which would be the network crate. In the network crate, we have a recv loop like:
```
pub fn recv(&mut self) {
    loop {
        let evt = self.rx.recv();
        // process it however
    }
}
```

fhaynes:
This provides a uniform way to send events from any thread to the network crate, and the logic to handle them in the network crate in the same. Sending to a channel does not block, so if the application slows down for some reason (heavy physics calcs, whatever), or the network crate slows down, the kernel won't start dropping the packets, they'll be buffered in the channel

fhaynes:
An effect of this method is that the mpsc channels will buffer forever (unless we explicitly limit the number)...from an OS perspective, that will cause RAM usage to balloon, and at some point the server will be killed by the OS.

PR #17, opened 2018-10-07
Add basic Event handling

LucioFranco:
This adds `Event` production to socket state. This is done via a mpsc
channel that can be consumed via `SocketState::events` which will
return a `Vec<Event>` to be consumed up the stack. This setup will
allow the `UdpSocket` to return this `Vec<Event>` back up to the user
to handle when the time is right in their implementation.

Other notable changes:
- Simple rustfmt on socket_state, udp, and events.
- Moved `Connected`, `Disconnected` and `TimedOut` to be a single
tuple of `Arc<RwLock<Connection>>` for clairty.
- Renamed `SocketState::create_connection_if_exists` to
`SocketState::get_connection_or_insert` also improved the way we use
locks in this. Since most calls to this function just needs a read
lock and not a write lock. A read lock is aquired initally to check if
there is a connection or not. If there is not, it will drop the read
lock and aquire a write lock to insert the new connection. At this
point it will also create a `Connected` event.

This PR is in reference to #11

### DIFF
--- a/src/net/socket_state.rs
+++ b/src/net/socket_state.rs
@@ -1,12 +1,14 @@
 use std::collections::HashMap;
+use std::sync::mpsc::{self, Receiver, Sender};
 use std::sync::{Arc, RwLock};
 use std::thread;
 use std::time::Duration;
 
-use packet::{Packet, PacketData};
-use packet::header::{FragmentHeader, PacketHeader};
-use net::{Connection,SocketAddr, NetworkConfig};
 use error::{NetworkError, Result};
+use events::Event;
+use net::{Connection, NetworkConfig, SocketAddr};
+use packet::header::{FragmentHeader, PacketHeader};
+use packet::{Packet, PacketData};
 use total_fragments_needed;
 
 // Type aliases
@@ -24,17 +26,22 @@ const TIMEOUT_POLL_INTERVAL: u64 = 1;
 pub struct SocketState {
     timeout: ConnectionTimeout,
     connections: ConnectionMap,
-    timeout_check_thread: thread::JoinHandle<()>
+    timeout_check_thread: thread::JoinHandle<()>,
+    events: (Sender<Event>, Receiver<Event>),
 }
 
 impl SocketState {
     pub fn new() -> Result<SocketState> {
         let connections: ConnectionMap = Arc::new(RwLock::new(HashMap::new()));
-        let thread_handle = SocketState::check_for_timeouts(connections.clone())?;
+        let (tx, rx) = mpsc::channel();
+
+        let thread_handle = SocketState::check_for_timeouts(connections.clone(), tx.clone())?;
+
         Ok(SocketState {
             connections: Arc::new(RwLock::new(HashMap::new())),
             timeout: TIMEOUT_DEFAULT,
-            timeout_check_thread: thread_handle
+            timeout_check_thread: thread_handle,
+            events: (tx, rx),
         })
     }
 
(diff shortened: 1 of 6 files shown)


src/net/socket_state.rs
```diff
@@ -81,15 +95,16 @@ impl SocketState {
...
+        } else {
             let num_fragments = total_fragments_needed(payload_length, config.fragment_size) as u8; /* safe cast max fragments is u8 */
 
             if num_fragments > config.max_fragments {
                 return Err(NetworkError::ExceededMaxFragments.into());
             }
 
             for fragment_id in 0..num_fragments {
-                let fragment = FragmentHeader::new(fragment_id, num_fragments, packet_header.clone());
+                let fragment =
```
TimonPost: Like it more in one line.

src/net/socket_state.rs
```diff
@@ -101,7 +116,8 @@ impl SocketState {
                 }
 
                 // get specific slice of data for fragment
-                let fragment_data = &payload[start_fragment_pos as usize..end_fragment_pos as usize]; /* upcast is safe */
+                let fragment_data =
+                    &payload[start_fragment_pos as usize..end_fragment_pos as usize]; /* upcast is safe */
```
TimonPost: Like it more in one line

src/net/socket_state.rs
```diff
@@ -309,14 +361,16 @@ mod test {
...
             }
         }
     }
 
-    fn simulate_packet_processing(data: Vec<u8>, config: &NetworkConfig) -> (SocketAddr, PacketData)
-    {
+    fn simulate_packet_processing(
+        data: Vec<u8>,
+        config: &NetworkConfig,
+    ) -> (SocketAddr, PacketData) {
```
TimonPost: Like it better in one line

src/net/socket_state.rs
```diff
@@ -309,14 +361,16 @@ mod test {
...
                 assert!(FragmentHeader::read(&mut cursor).is_ok())
             }
         }
     }
 
-    fn simulate_packet_processing(data: Vec<u8>, config: &NetworkConfig) -> (SocketAddr, PacketData)
-    {
+    fn simulate_packet_processing(
+        data: Vec<u8>,
+        config: &NetworkConfig,
```
TimonPost: One line much better

src/net/socket_state.rs
```diff
@@ -222,7 +269,7 @@ mod test {
         let addr = format!("{}:{}", TEST_HOST_IP, TEST_PORT).to_socket_addrs();
```
fhaynes: Have we settled on parse() vs to_socket_addrs()?

src/net/udp.rs
```diff
@@ -1,9 +1,9 @@
...
-
-use packet::{Packet, PacketProcessor};
-use net::{NetworkConfig, SocketState};
 use error::{NetworkError, Result};
+use events::Event;
+use net::{NetworkConfig, SocketState};
+use packet::{Packet, PacketProcessor};
 
 /// Maximum amount of data that we can read from a datagram
 const BUFFER_SIZE: usize = 1024;
```
fhaynes: We don't need to do it now, but we should probably make an issue to make this configurable.

src/net/udp.rs
```diff
@@ -84,8 +91,8 @@ mod test {
     #[test]
     #[ignore]
```
fhaynes: Why is this test ignored?

src/net/udp.rs
```diff
@@ -84,8 +91,8 @@ mod test {
...
     fn send_receive_1_pckt() {
-        let mut send_socket = UdpSocket::bind("127.0.0.1:12347",NetworkConfig::default()).unwrap();
-        let mut recv_socket = UdpSocket::bind("127.0.0.1:12348",NetworkConfig::default()).unwrap();
+        let mut send_socket = UdpSocket::bind("127.0.0.1:12347", NetworkConfig::default()).unwrap();
+        let mut recv_socket = UdpSocket::bind("127.0.0.1:12348", NetworkConfig::default()).unwrap();
 
         let addr = "127.0.0.1:12348".parse().unwrap();
```
fhaynes: Should we change to to_socket_addr?

src/net/udp.rs
```diff
@@ -107,28 +114,28 @@ mod test {
     #[test]
     #[ignore]
```
fhaynes: Also this test...why the ignore?

src/net/socket_state.rs
```diff
@@ -44,14 +51,21 @@ impl SocketState {
...
+                packet.payload().len(),
+                config.max_packet_size
+            );
             return Err(NetworkError::ExceededMaxPacketSize.into());
         }
 
-        let connection = self.create_connection_if_not_exists(&packet.addr())?;
+        let connection = self.get_connection_or_insert(&packet.addr())?;
 
         let mut connection_seq: u16 = 0;
```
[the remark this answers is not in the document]
TimonPost: Yea think that would be good. Because this function returns out of my head Arc<RwLock<Connection>>

src/net/udp.rs
TimonPost: It was they are working now.

fhaynes:
I'm good as soon as all the Issues are made, then I think we can merge.

LucioFranco:
bors r=fhaynes,TimonPost

PR #18, opened 2018-10-07
Implementing an connection pool for udp virtual connections.

TimonPost:
So in discord we spoke about refactoring `SocketState`.

I have moved all the connection logic from `SocketState` type to its own type. Now we have an `ConnectionPool` which will manage the virtual connections. 

So why this you may ask. Because we want to separate concerns here. And having connection logic appart from receving logic seems nice to me. And a nice beginning for refactoring `SocketState.`

I can make the connection pool generic to also support TCP connections but did not know if that was necessarily since TCP connections work differently.


src/net/connection/connection_pool.rs
```diff
@@ -0,0 +1,95 @@
...
+                            error!("Unable to acquire read lock to check for timed out connections")
+                        }
+                    }
+                    thread::sleep(poll_interval);
+                })?
+        )
+    }
+
+    /// Check if there are any connections that have not been active for the given Duration.
+    fn check_for_timeouts(connections: &mut HashMap<SocketAddr, Arc<RwLock<VirtualConnection>>>, sleepy_time: Duration) {
```
fhaynes: Type alias here?

src/net/connection/connection_pool.rs
```diff
@@ -0,0 +1,95 @@
...
+        for (key, value) in connections.iter() {
+            if let Ok(c) = value.read() {
+                if c.last_heard() >= sleepy_time {
+
+                    // TODO: pass up client TimedOut event
```
fhaynes: Will this be in another PR?

src/net/socket_state.rs
```diff
@@ -118,7 +107,7 @@ impl SocketState {
 
     /// This will return all dropped packets from this connection.
     pub fn dropped_packets(&mut self, addr: SocketAddr) -> Result<Vec<Packet>> {
-        let connection = self.create_connection_if_not_exists(&addr)?;
+        let connection = self.connections.insert_if_not_exists(&addr)?;
         let mut lock = connection
             .write()
             .map_err(|_| NetworkError::AddConnectionToManagerFailed)?;
```
fhaynes: Should we bubble up this error?

src/net/socket_state.rs
```diff
@@ -129,7 +118,7 @@ impl SocketState {
 
     /// This will process an incoming packet and update acknowledgement information.
     pub fn process_received(&mut self, addr: SocketAddr, packet: &PacketHeader) -> Result<()>{
-        let connection = self.create_connection_if_not_exists(&addr)?;
+        let connection = self.connections.insert_if_not_exists(&addr)?;
         let mut lock = connection
             .write()
             .map_err(|_| NetworkError::AddConnectionToManagerFailed)?;
```
fhaynes: Should we bubble up this error?

src/net/connection/connection_pool.rs
```diff
@@ -0,0 +1,95 @@
...
+    pub fn new() -> ConnectionPool
+    {
+        let sleepy_time = Duration::from_secs(TIMEOUT_DEFAULT);
+        let poll_interval = Duration::from_secs(TIMEOUT_POLL_INTERVAL);
+
+        ConnectionPool { timeout: TIMEOUT_DEFAULT, connections: Arc::new(RwLock::new(HashMap::new())), sleepy_time, poll_interval }
+    }
+
+    /// Set the timeout before an client will be seen as disconnected.
+    pub fn set_timeout(&mut self, timeout: ConnectionTimeout)
```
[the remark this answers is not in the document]
LucioFranco: Why not just accept a `Duration`?

src/net/connection/connection_pool.rs
```diff
@@ -0,0 +1,95 @@
...
+    {
+        self.timeout = timeout;
+    }
+
+    /// Insert connection if it does not exists.
+    pub fn insert_if_not_exists(&mut self, addr: &SocketAddr) -> Result<Arc<RwLock<VirtualConnection>>> {
```
LucioFranco: Please rename this to `get_connection_or_insert()` and use https://github.com/amethyst/laminar/blob/master/src/net/socket_state.rs#L220 as an example to implement this with proper read and write lock usage.

src/net/connection/connection_pool.rs
```diff
@@ -0,0 +1,95 @@
...
+            .map_err(|_| NetworkError::AddConnectionToManagerFailed)?;
+
+        let connection = lock
+            .entry(*addr)
+            .or_insert_with(|| Arc::new(RwLock::new(VirtualConnection::new(*addr))));
+
+        Ok(connection.clone())
+    }
+
+    /// Start loop that will detect if connections will are disconnected.
```
LucioFranco: Remove the `will`

src/net/connection/connection_pool.rs
```diff
@@ -0,0 +1,95 @@
...
+    ///
+    /// This function starts a background thread that does the following:
+    /// 1. Gets a read lock on the HashMap containing all the connections
+    /// 2. Iterate through each one
+    /// 3. Check if the last time we have heard from them (received a packet from them) is greater than the amount of time considered to be a timeout
```
LucioFranco: You should add a module doc that explains how timeouts work. This would be something useful. 

If not in here, somewhere else would be helpful.

src/net/connection/connection_pool.rs
```diff
@@ -0,0 +1,95 @@
...
+        let connections = self.connections.clone();
+        let poll_interval = self.poll_interval;
+
+        Ok(thread::Builder::new()
+            .name("check_for_timeouts".into())
+            .spawn(move ||
+                loop {
+                    match connections.write() {
+                        Ok(mut lock) => {
+                            Self::check_for_timeouts(&mut *lock, poll_interval);
```
LucioFranco: I think it would be better to use `ConnectionPool::` here maybe, just to be that even more so clear.

src/net/connection/connection_pool.rs
```diff
@@ -0,0 +1,95 @@
...
+    pub fn start_time_out_loop(&self) -> Result<thread::JoinHandle<()>>
+    {
+        let connections = self.connections.clone();
+        let poll_interval = self.poll_interval;
+
+        Ok(thread::Builder::new()
+            .name("check_for_timeouts".into())
+            .spawn(move ||
+                loop {
+                    match connections.write() {
```
LucioFranco: I don't think you need a `write` lock here? Since, you are just reading from the hashmap?

src/net/connection/connection_pool.rs
LucioFranco: This is implemented in master, so you should rebase and add a `send` call here.

src/net/connection/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+
+mod virtual_connection;
+mod connection_pool;
+mod quality;
+
+pub use self::virtual_connection::VirtualConnection;
+pub use self::connection_pool::ConnectionPool;
+pub use self::quality::NetworkQuality;
+
+pub type ConnectionTimeout = u64;
```
LucioFranco: I think this should be a `Duration` for this.

src/net/connection/virtual_connection.rs
```diff
@@ -1,31 +1,31 @@
...
+use net::{ExternalAcks, LocalAckRecord, NetworkQuality};
 use packet::Packet;
 use std::fmt;
 use std::net::SocketAddr;
 use std::time::{Duration, Instant};
 
 /// Contains the information about a certain 'virtual connection' over udp.
 /// This stores information about the last sequence number, dropped packages, packages waiting for acknowledgement and acknowledgements gotten from the other side.
-pub struct Connection {
+pub struct VirtualConnection {
```
LucioFranco: Why the rename?

src/net/connection/connection_pool.rs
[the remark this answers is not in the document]
TimonPost: Yea lets do that.

src/net/connection/connection_pool.rs
TimonPost: I am going to rebase on @LucioFranco and probably after that I have to move the event logic to the connection pool as well.

src/net/socket_state.rs
TimonPost: No.

src/net/connection/connection_pool.rs
TimonPost: Let move that to another PR the connection timeout stuff is mostly from @fhaynes. I think it will be genearaly a good idea to make some docs.

src/net/connection/virtual_connection.rs
TimonPost: Because udp connection is not an connection it is an virtual connection not one that is really there.  gaffer on games call it also like that. So I thought if we follow his blogs then lets keep the names the same also.

fhaynes: Yeah, I agree on that. It's an important distinction at this level.

LucioFranco: Works for me! Thanks.

TimonPost:
Going to reopen but due some conflict I need to remove this branch

PR #21, opened 2018-10-08
Implemented Connection Pool.

TimonPost:
So in discord we spoke about refactoring `SocketState`.

I have moved all the connection logic from `SocketState` type to its own type. Now we have an `ConnectionPool` which will manage the virtual connections. 

So why this you may ask. Because we want to separate concerns here. And having connection logic appart from receving logic seems nice to me. And a nice beginning for refactoring `SocketState.`

I can make the connection pool generic to also support TCP connections but did not know if that was necessarily since TCP connections work differently.

### DIFF
--- a/src/events.rs
+++ b/src/events.rs
@@ -1,28 +1,28 @@
 use std::sync::{Arc, RwLock};
 
-use net::{Connection, Quality};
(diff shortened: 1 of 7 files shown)


LucioFranco:
Good work!
bors r+

PR #22, opened 2018-10-09
Benchmarks and Integration tests

TimonPost:
I have added benchmarks for processing a packet when received or before send. I did not use the UDP socket for this but instead used directly the processing code. I did this since we don't want to benchmark receiving and sending over a socket. But instead the processing logic.

I have also added a few integration tests.


tests/common.rs
```diff
@@ -0,0 +1,105 @@
...
+                            Err(_) => {}
+                        }
+                    }
+                }
+
+                if second_counter.elapsed().as_secs() >= 1 {
+                    // reset counter
+                    second_counter = Instant::now();
+
+//                    println!("Packet throughput: {} p/s", packet_throughput);
```
fhaynes: This can probably be removed

tests/common.rs
```diff
@@ -0,0 +1,105 @@
...
+
+            let len = data_to_send.len();
+
+            for i in 0..packets_to_send {
+                let send_result = client.send(Packet::new(host, data_to_send.clone()));
+
+                if len <= config.fragment_size as usize {
+                    assert_eq!(send_result.unwrap(), len + constants::PACKET_HEADER_SIZE as usize);
+                }else {
+                    // if fragment, todo: add size assert.
```
fhaynes: Need an issue for this

tests/fragments.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    server.add_client(test_data.to_vec(), client).join();
+
+    // notify server to stop receiving.
+    tx.send(true);
+
+    let total_received = server_thread.join().unwrap();
+    let elapsed_time = stopwatch.elapsed();
+
+//    println!("Total Duration: {:?}", elapsed_time);
```
fhaynes: Can we remove these?

tests/multiple_clients.rs
```diff
@@ -0,0 +1,66 @@
...
+        handle.join();
+    }
+
+    // notify server to stop receiving.
+    tx.send(true);
+
+    let total_received = server_thread.join().unwrap();
+    let elapsed_time = stopwatch.elapsed();
+
+//    println!("Total Duration: {:?}", elapsed_time);
```
fhaynes: Can we remove these?

tests/normal.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    server.add_client(test_data.to_vec(), client).join();
+
+    // notify server to stop receiving.
+    tx.send(true);
+
+    let total_received = server_thread.join().unwrap();
+    let elapsed_time = stopwatch.elapsed();
+
+//    println!("Total Duration: {:?}", elapsed_time);
```
fhaynes: Can we remove these?

PR #30, opened 2018-10-13
RTT estimation implementation.

TimonPost:
This is a basic implementation of congestion avoidance. 

During development, I stumbled upon a few things of where I am not sure of on how to go further with it. But first, let's go into detail what this PR code does.

**Round Trip Time (RTT)**
The time between you sending the packet and you receiving an acknowledgement from the other side is called RTT. To avoid congestion we first need to find a way to calculate the `RTT` value of our connection so we can decide on top of that value if we have bad or good internet speeds.

_Smoothing factor_
So you could say: "very simple, measure the time between sending and receiving you got the `RTT` and you're done right?" No! This is because a packet can travel any path over the internet the `RTT` can always defer every time you calculate it. And imagine a short internet lag we will directly get a huge RTT back. So we need to smooth out that RTT factor by some amount. Gaffer says that 10% of the RTT will be just fine. With this smoothed RTT we will be able to add it to our current RTT. 

_Allowed RTT value_
So now we have the smoothed RTT and or current RTT, GREAT! But RTT on its own is not bad. So there may be some max allowed RTT. We need to subtract that amount from our measured RTT multiplied by the smoothing factor. 

The formule would look like the following:

```
// rtt_max_value is in ms
// rtt_smoothing_factor is in %
let new_rtt_value = (rtt - rtt_max_value) * rtt_smoothing_factor.
```
Lets look at an example with numbers. The RTT values are in milliseconds.

_bad internet_
```
// this will result into: 5
let new_rtt_value = (300 - 250) * 0.10.
```

_good internet_
```
// this will result into: -15
let new_rtt_value = (100 - 250) * 0.10.
```

As you see when or calculation is under 250ms we get a negative result, which is in this case positive. When or calculation is above 250ms it will be positive, which is in this case negative.

So each time we receive an acknowledgement we can add our result, of the above formula, to the RTT time saved in the connection. Note that the above is implemented now.

**Problem** 
So in our situation, we don't have acknowledgement packets but instead, when we receive a packet from the other side we check on which packets the other side had received from us. So we only know if the other side has received or packets when we have gotten a packet from the other side.

Now if the other side is some kind of idle and or isn't sending any data for let's say 400ms our RTT formula and calculations will take the time when the packet is received. And we get a high RTT which we can interpreter as bad network quality. However, in this case, it isn't, the client was just idling.

**Question**
How are going to handle this? 

I have to things to add to this question:
1. Should we need to send ack packets? 
2. Should we accept the fact that idling clients would not occur, because most of the time the client will send packets at the rate of 30-60 p/s. 

The current situation will work but only if the client is sending at some interval n packets. To my knowledge, this wouldn't be any problem because we are building this UDP-based protocol to support high traffic.


fhaynes:
For idle clients, we should use a heartbeat that is every 1-5s or so

src/net/connection/quality.rs
```diff
@@ -1,18 +1,93 @@
...
+impl NetworkQualityMeasurer
+{
+    pub fn new(config: NetworkConfig) -> NetworkQualityMeasurer
+    {
+        NetworkQualityMeasurer { config }
+    }
+
+    /// This will calculate the round trip time (rtt) from the given acknowledgement.
+    /// Where after it updates the rtt from the given connection.
+    pub fn update_connection_rtt(&self, connection: &mut RwLockWriteGuard<VirtualConnection>, ack_seq: u16)
```
fhaynes: Huh, this is interesting. Just out of curiosity, why not pass in an arc?

src/net/connection/quality.rs
```diff
@@ -1,18 +1,93 @@
...
+    }
+
+    /// Converts an duration to milliseconds.
+    ///
+    /// `as_milliseconds` is not supported yet supported in rust stable.
+    /// See this stackoverflow post for more info: https://stackoverflow.com/questions/36816072/how-do-i-get-a-duration-as-a-number-of-milliseconds-in-rust
+    fn as_milliseconds(&self, duration: Duration) -> u64
+    {
+        let nanos = duration.subsec_nanos() as u64;
+        (1000 * 1000 * 1000 * duration.as_secs() + nanos) / (1000 * 1000)
```
fhaynes: I feel like this could be made more concise, but not sure how, and not sure if it is worth it.

src/net/connection/quality.rs
```diff
@@ -25,4 +100,31 @@ mod test {
...
+        let mut config = NetworkConfig::default();
+        // for test purpose make sure we set smoothing factor to 10%.
+        config.rtt_smoothing_factor = 0.10;
+        config.rtt_max_value = 250;
+
+        let network_quality = NetworkQualityMeasurer::new(config.clone());
+        let smoothed_rtt = network_quality.smooth_out_rtt(300);
+
+        // 300ms has exceeded 50ms over the max allowed rtt. So we check if or smoothing factor is now 10% from 50.
+        assert_eq!(smoothed_rtt, 5.);
```
fhaynes: I assume 5. is equivalent to 5.0? If so, can we make this 5.0?

src/net/connection/virtual_connection.rs
```diff
@@ -27,6 +29,8 @@ impl VirtualConnection {
             last_heard: Instant::now(),
             quality: NetworkQuality::Good,
             remote_address: addr,
+            congestion_avoidance_buffer: FragmentBuffer::with_capacity(<u16>::max_value() as usize),
+            rtt: 0.,
```
fhaynes: Same here about the 0. vs 0.0

tests/common.rs
```diff
@@ -55,7 +57,7 @@ impl ServerMoq {
                     // reset counter
                     second_counter = Instant::now();
 
-//                    println!("Packet throughput: {} p/s", packet_throughput);
+                    // println!("Packet throughput: {} p/s", packet_throughput);
```
fhaynes: can we remove commented out code?

src/net/connection/quality.rs
TimonPost: I am also not sure how. Rust will at one-time support converting duration to milliseconds but for now, I had to copy some code from StackOverflow, see the link in the comments.

src/net/connection/quality.rs
TimonPost: Yes, it will, and yes I can change that.

src/net/connection/quality.rs
TimonPost: This was because this function gets called when the lock is already requested on the connection. So I thought why not directly pass in an unlocked connection into this function.

fhaynes: Yeah, that's pretty cool. I've never thought to do it that way. Nice.

fhaynes:
@TimonPost I made an attempt at resolving the conflict. Let me know if things still look good to you and if so we can merge.

TimonPost:
bors r=fhaynes

TimonPost:
bors r=fhaynes

PR #34, opened 2018-10-14
Basic Md Book setup

TimonPost:
Created a basic setup for the md book. 

Maybe better to don't review this and just merge this pr. It is probably much faster if another PR fixes all issues etc. there are now. I won't say that this book is perfect, but it is at least a basic implementation on which could be iterated.


TimonPost:
bors -r=fhaynes

TimonPost:
bors r=fhaynes

LucioFranco:
This is great! Thanks.

PR #35, opened 2018-10-15
Refactor 1: Moved some common logic to its own folder

TimonPost:
This PR is part of a series of PR's that will refactor this crate. 

This PR contains:
- Moving some in common having logic to its own folder. 
  This is done because `CongestionData` an `ReasemblyData` both were entries in the `FragmentBuffer` probably in the future we are going to have more types of entries.  So I thought let's move this `FragmentBuffer` and all the entries to its own folder out of the folder `packet`, because these are not packet related an should be placed in an different folder.
- Renamed `FragmentBuffer` to `SecuenceBuffer`
  I have done this because temporary fragments are stored with this buffer and temporary congestion avoidance data is stored in this buffer by sequence number.


TimonPost:
bors r=fhaynes

TimonPost:
bors r=fhaynes

TimonPost:
bors r=fhaynes

Issue #39, opened 2018-10-16
Thinking of packet processing logic.


TimonPost:
I am thinking of the packet workflow under which you can think: easiest way for creating packets and sending them over to the other side with all the options that a packet could have like encryption etc.

Take the following things into consideration:
1. Options like encryption/decryption
2. Options for different reliability strategies: Unreliable, ReliableUnordered, Sequenced ReliableOrdered.
3. Must be configurable.
4. Must be easily extendable.
5. The user does not have to be bothered with all the different network tactics and implementations on how to send packets. The only thing a user should care about is, sending bytes!

Here I have a draft on how to separate packets of different types. And on how we can convert a simple user packet to some more complex packet types about which the user should not care about.

https://gist.github.com/TimonPost/303c53fa1454d724120a1173b04a9745

Now working on processing the bytes of each packet this issue will be updated soon.

OvermindDL1:
In addition to priorities to determine which to focus on sending first, and also channels so you can have multiple 'ordered streams' in progress at the same time.

TimonPost:
There was once some kind of implementation of channels. But some folks did not like opening different UPD sockets for different kind of implementations. See [this](https://github.com/LucioFranco/amethyst-protocol/pull/21) PR

OvermindDL1:
> But some folks did not like opening different UPD sockets for different kind of implementations.

Because you shouldn't need to.  It should be part of the application packet header, using a single byte is well worth channel support for 255 different channels for anything sequenced or ordered (and unordered doesn't need it at all).

TimonPost:
I like the channel idea but am not sure if I understand it right. Can you be more specific with the channels a bit? 

Currently, how I am imagine this is having channels who will process each packet on its own way. Based on the id of the packet it will be processed by a certain channel. Then my next questions are:

1. How is the user able to get the packet from the channel. 
2. Does the user have access to the channel contents? And processing stuff.
3. How does the user get the newest packets from the channels? 
4. Do you have some example of the channels Raknet is using? I have been wandering around in their code but not have seen anything related to channels.

OvermindDL1:
In the Raknet docs it speaks of the costs (my bolding):  http://www.raknet.net/raknet/manual/systemoverview.html

* Per datagram:
  1. 1 byte for bitflags
  2. 4 bytes for timestamp, used to calculate RTT for congestion control. RakNet version 4 does not transmit this field.
  3. 3 bytes for sequence number, used to lookup datagrams for ACKs

[...]

PR #42, opened 2018-10-17
Added support for sending different kinds packets

TimonPost:
This PR is part of issue #39 and allows having different packet types and is not about processing and interpreting different packets. The next PR I am going to make will be about that as described in #39 

This gist is some  basic draft: https://gist.github.com/TimonPost/303c53fa1454d724120a1173b04a9745

Don't mind if there aren't any comments yet. I forgot to comment some things so that will be added later.

As you can see I have used "sequenced unreliable" packet for now because this is the type we provide now. But later with the next PR this will be dynamic and will be based on packet arrived.

### DIFF
--- a/src/infrastructure/delivery_method.rs
+++ b/src/infrastructure/delivery_method.rs
@@ -1,0 +1,84 @@
+/// This enum defines on how an packet would be delivered to the other side.
+#[derive(Copy, Clone, Debug,PartialOrd, PartialEq, Eq)]
+pub enum DeliveryMethod
+{
+    /// Unreliable. Packets can be dropped, duplicated or arrive without order
+    ///
+    /// *Details*
+    ///
+    ///  1. Unreliable
+    ///  2. No guarantee for delivery.
+    ///  3. No guarantee for order.
+    ///  4. No way of getting dropped packet
+    ///  5. Duplication possible
+    ///
+    /// Basically just bare UDP
+    Unreliable,
+    /// Reliable. All packets will be sent and received, but without order
+    ///
+    /// *Details*
+    ///
+    ///  1. Reliable.
+    ///  2. Guarantee of delivery.
+    ///  3. No guarantee for order.
+    ///  4. Packets will not be dropped.
+    ///  5. Duplication not possible
+    ///
+    /// Basically this is almost TCP like without ordering of packets.
+    ReliableUnordered,
+    /// Unreliable. Packets can be dropped, but never duplicated and arrive in order
+    ///
+    /// This will create an reliable ordered packet.
+    ///
+    /// *Details*
+    ///
+    ///  1. Unreliable.
+    ///  2. No guarantee of delivery.
+    ///  3. Guarantee for order.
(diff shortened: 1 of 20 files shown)


src/infrastructure/delivery_method.rs
```diff
@@ -0,0 +1,38 @@
...
+    SequencedOrdered,
+    /// Unreliable. Packets can be dropped, and arrive out of order but you will be able to retrieve dropped packet.
+    SequencedUnordered,
+    /// Reliable and ordered. All packets will be sent and received in order
+    ReliableOrdered,
+}
+
+impl DeliveryMethod {
+    pub fn get_delivery_method_id(delivery_method: DeliveryMethod) -> u8 {
+        match delivery_method {
```
LucioFranco: you should be able to do `delivery_method as u8`
https://doc.rust-lang.org/beta/reference/items/enumerations.html#custom-discriminant-values-for-field-less-enumerations

src/infrastructure/delivery_method.rs
```diff
@@ -0,0 +1,38 @@
...
+            DeliveryMethod::Unreliable => 0,
+            DeliveryMethod::ReliableUnordered => 1,
+            DeliveryMethod::SequencedOrdered => 2,
+            DeliveryMethod::SequencedUnordered => 3,
+            DeliveryMethod::ReliableOrdered => 4,
+        }
+    }
+
+    pub fn get_delivery_method_from_id(delivery_method_id: u8) -> DeliveryMethod {
+        match delivery_method_id {
```
LucioFranco: and I think you can do the reverse

src/packet/header/packet.rs
```diff
@@ -47,7 +54,8 @@ impl HeaderParser for PacketHeader {
 
     fn parse(&self) -> <Self as HeaderParser>::Output {
         let mut wtr = Vec::new();
-        wtr.write_u8(0)?;
+        wtr.write_u8(PacketTypeId::get_id(self.packet_type_id))?;
+        wtr.write_u8( DeliveryMethod::get_delivery_method_id(self.delivery_method))?;
```
LucioFranco: extra space here :)

src/infrastructure/delivery_method.rs
TimonPost: Thanks, of course, knew that.

src/net/constants.rs
```diff
@@ -9,4 +11,4 @@ pub const FRAGMENT_SIZE_DEFAULT: u16 = 1024;
 ///
 /// This is not strictly guaranteed -- there may be less room in an ethernet frame than this due to
 /// variability in ipv6 header size.
```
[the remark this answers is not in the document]
TimonPost: ipv6 because we should work from the biggest values. ipv4 is just 20 bytes so we are safer when we have  an MTU with more space by calculating with the size of IPV6 wich is 40 bytes.

TimonPost:
bors r=fhaynes

PR #43, opened 2018-10-17
Added in a binary for testing

fhaynes:
This is a start on a binary that can function as a simple client or server for more complex testing scenarios.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ byteorder = "1.2.6"
 log = "0.4"
 failure = "0.1"
 failure_derive = "0.1"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -27,10 +27,14 @@ byteorder = "1.2.6"
 log = "0.4"
 failure = "0.1"
 failure_derive = "0.1"
+clap = { version = "2.32", features = ["yaml"] }
```
LucioFranco: should this not be added to `[dev-dependencies]` or even maybe a better field. Since we are a library not an application.

fhaynes: We are both

Issue #45, opened 2018-10-18
Guidelines, overal crate improvements.


TimonPost:
Here are a couple of things I think who are important for a library like this one. If you see something lac of these issues feel free to open a PR. When I read some code from other libs like this one, but in C++/C I saw that there was a huge shortcoming on most of the following points. Which makes the code difficult to read and understand.

- Every type, method and some fields should be commented.
- Calculations should be documented. Whether it would be in a PR or code. But it must be clear what is done. 
- Don't hard code values. We should only use config, consts etc hardcoded values are the root of all evil.
- Everything should be tested, so feel free to write tests.
- Documentation.
- Keep files small better have small files with small pieces of logic than having one file with 1000 lines of logic with multiple types/structs etc. Note that I speak of logic, tests not included.
- Absolutely no panics/unwraps in the logic we need to have propper error handling. You may have panics/unwraps in test code.
- Grammar / English improvement. I am not a native English speaker so forgive me when making some mistakes. I am trying to avoid making them in the first place. But some checks could not be wrong.

Xaeroxe:
Over on the main Amethyst repo we implemented a different approach, which was to use warnings most of the time except for on CI, where we added flags to cargo to deny on warnings.  This allows you to prototype quickly, but still quality gates the repo.  See https://github.com/amethyst/amethyst/pull/1011 and https://github.com/amethyst/amethyst/pull/1025 for more.

LucioFranco:
@Xaeroxe I like this approach. And we should be following what amethyst/amethyst does anyways since we are another repo under the same org.

CBenoit:
Maybe we should add rustfmt verification in CI to make sure formatting is uniform.

LucioFranco:
@CBenoit yes, this is the plan. Mostly all of the amethyst crates are waiting for `rustfmt` to `1.0` to formalize that. They are like a month overdue 😄

LucioFranco:
punting this to `0.2.0`

fhaynes:
I'm going to try to add code coverage for 0.1.0 as part of this.

TimonPost:
Isn't really needed I think. We could better put something like this in /docs. Which it already does over here: https://github.com/amethyst/laminar/blob/master/docs/CONTRIBUTING.md

Issue #31, opened 2018-10-13
Clean up code base


LucioFranco:
Steps:
- [x] Run Clippy and fix suggestions
- [x] Clean up warnings in the crate, test, and examples
- [x] Rustfmt the repo with `v0.99`

TimonPost:
Updates #31 en #32

CBenoit:
I already mentioned it here: https://github.com/amethyst/laminar/pull/44
But I feel like `packet/header/packet.rs` should be renamed to `packet/header/packet_header.rs`.
I consider it to be clearer:
- I regularly open the wrong file because it is named like `packet/packet.rs`
- It doesn't contains Packet but PacketHeader. When I read the codebase the first time it was confusing.

TimonPost:
Do it, I first called that way since it was in the header folder so thought it was unnecessarily to add `_header` to it. But I find it confusing to so i'll be fine with it.

TimonPost:
This is an ongoing process but for now we did #36 , #57

PR #48, opened 2018-10-19
Refactor 3: fix easy warnings

CBenoit:
Related to https://github.com/amethyst/laminar/issues/31

I fixed most easy warnings:
- unused import
- unused value
- useless mut
- some unused variables

I also fixed the overflowing_literals warnings, you might want to have a closer look at `sequence_buffer.rs` file diff.

Also, there were missing `#[cfg(test)]` causing unused import warnings when not building tests.

I didn't fix warnings related to Error handling because these may require special attention.

Also there are some unused parameters and unused struct attributes that I believe will be used in the future.

Some remarks:
- Purpose of this attribute is not documented + the attribute itself is never used.
 https://github.com/amethyst/laminar/blob/master/src/net/connection/connection_pool.rs#L23
- These are arbitrarily hardcoded values.
 https://github.com/amethyst/laminar/blob/master/src/net/connection/connection_pool.rs#L29
 https://github.com/amethyst/laminar/blob/master/src/net/connection/connection_pool.rs#L33
 Shall I create constants for them?
- This is unused private method
 https://github.com/amethyst/laminar/blob/master/src/net/socket_state.rs#L174
 Shall we delete it or make it public?

There is a lot of dead code. This is not a good practice (related to https://github.com/amethyst/laminar/issues/45 )

### DIFF
--- a/src/net/connection/connection_pool.rs
+++ b/src/net/connection/connection_pool.rs
@@ -56,7 +56,7 @@ impl ConnectionPool {
         Ok(connection.clone())
     }
 
-    /// Start loop that will detect if connections will are disconnected.
+    /// Start loop that detect connection timeouts.
     ///
     /// This function starts a background thread that does the following:
     /// 1. Gets a read lock on the HashMap containing all the connections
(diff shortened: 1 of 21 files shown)


src/net/connection/connection_pool.rs
```diff
@@ -56,7 +56,7 @@ impl ConnectionPool {
         Ok(connection.clone())
     }
 
-    /// Start loop that will detect if connections will are disconnected.
+    /// Start loop that detect connection timeouts.
```
LucioFranco: `Start a loop that detects when a connection has timed out.`

CBenoit:
Should be fine now :+1:

LucioFranco:
Cool thanks

bors r+

Issue #49, opened 2018-10-19
Rethinking error handling


TimonPost:
We need to rethink how or errors are handled. I have been researching and came to the conclusion that we should use this way of error handling: https://boats.gitlab.io/failure/error-errorkind.html.

I have made a list of all error messages and divided them into groups.

```rust
/// Most high-level error type that contains network errors 
enum NetworkError
{
    #[fail(display = "Something went wrong when sending")]
    SendingError { inner: SendingError },
    #[fail(display = "Something went wrong when receiving")]
    ReceivingError { inner: ReceivingError },
    #[fail(display = "Something went wrong when perfoming action on socket")]
    SocketError { inner: SocketError },
    #[fail(display = "Lock poisoned")]
    FailedToAddConnection { error: ::std::sync::PoisonError }
}

/// Socket errors that could occur with the socket.
enum SocketError
{
    #[fail(display = "Unable to set nonblocking option")]
    UnableToSetNonblocking,
    #[fail(display = "Unable to create UDP SocketState structure")]
    UDPSocketStateCreationFailed,
}

/// Errors that could occur when sending an packet.
enum SendingError
{
    #[fail(display = "Io operation failed")]
    IoError { io_error: io::Error },
    #[fail(display = "Something went wrong with fragmentation")]
    FragmentError { inner: FragmentError },
    #[fail(display = "Something went wrong with constructing/parsing packets")]
    PacketError { inner: PacketError },
}

/// Errors that could occur when receiving
enum ReceivingError
{
    #[fail(display = "Io operation failed")]
    IoError { io_error: io::Error },
    #[fail(display = "Something went wrong with fragmentation")]
    FragmentError { inner: FragmentError },
    #[fail(display = "Something went wrong with receiving/parsing packets")]
    PacketError { inner: PacketError },
}

/// Errors that could occur with constructing parsing packet contents
enum PacketError {
    #[fail(display = "The packet size was bigger than the max allowed size.")]
    ExceededMaxPacketSize,
    #[fail(display = "Something went wrong when parsing the packet header")]
    ParsingError { io_error: io::Error }
}

/// Errors that could occur with constructing parsing fragment contents
enum FragmentError
{
    #[fail(display = "No packet header attached to fragment.")]
    FragmentNotFound,
    #[fail(display = "The total of fragments the packet was divided into is bigger than the allowed fragments.")]
    ExceededMaxFragments,
    #[fail(display = "The fragment processed was already received.")]
    AlreadyProcessedFragment,
[...]

PR #51, opened 2018-10-19
Improved error handling with more detailed error messages.

TimonPost:
Improved error handling. 

Things changed:
- Added some `ErrorKinds` which contains some group related error messages. 
- Added more detailed `fail()` descriptions with inner type in a description.
- Added some basic `From` implementations who are making it easier to return errors from anywhere.
- Moved error handling to its own folder.

Related to this issue: #49 

**_Special note why big PR_**
Because error handling is all over the system almost in every file. That is why there are so many files changed. I could not think of a way to split it into pieces. But if you look at LoC, it is not that much :)


src/packet/raw_packet_data.rs
```diff
@@ -11,7 +11,7 @@ pub struct RawPacketData {
 }
 
 impl RawPacketData {
-    pub fn new(header: &HeaderParser<Output = Result<Vec<u8>>>, body: Vec<u8>) -> RawPacketData {
+    pub fn new(header: &HeaderParser<Output = NetworkResult<Vec<u8>>>, body: Vec<u8>) -> RawPacketData {
```
fhaynes: Missing docstring

src/error/error_kinds.rs
```diff
@@ -0,0 +1,43 @@
+use std::io;
+
+/// Errors that could occur with constructing parsing packet contents
+#[derive(Fail, Debug, PartialEq, Eq, Clone)]
+pub enum PacketErrorKind {
+    #[fail(display = "Could not process the packet, because the packet size was bigger than the max allowed size.")]
```
LucioFranco: I would remove the `could not process the packet, because` because its extra and its repeated. I would just have it say "The packet size was bigger than the max allowed size.`

src/error/network_error.rs
```diff
@@ -0,0 +1,115 @@
...
+use std::fmt::{self, Display,Formatter};
+use std::io;
+
+use failure::{Fail, Backtrace, Context};
+
+#[derive(Fail, Debug)]
+/// Enum with all possible network errors that could occur.
+pub enum NetworkErrorKind
+{
+    #[fail(display = "Something went wrong with receiving/parsing fragments, see the inner error for more information. Reason: {:?}.", inner)]
```
LucioFranco: Same here lets keep these short, and just return the inner error.

TimonPost:
bors r=fhaynes, TimonPost

PR #55, opened 2018-10-20
Implemented protocol id for headers.

TimonPost:
Implemented protocol id see #4  for more information.

Please during review check these out:
- laminar-0.1.0 
From this value or crc32 will be generated do we agree on this notation for protocol version.
- I use unsafe in on place see /src/protocol_version.rs, I can change this to `lazy_static` but like the current way, it could also be done. What should I do with this? 
- I use CRC32 for encoding of protocol version. Maybe an option to have CRC16? I don't know if there is some major difference between the who other than the size.

O shit. I see that I have to comment some things from `ProtocolVersion` I'll update it later.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,8 @@ failure = "0.1"
 failure_derive = "0.1"
 clap = { version = "2.32", features = ["yaml"], optional = true }
 env_logger = { version = "0.5.13", optional = true }
+crc = "1.8"
+lazy_static = "1.1.0"
 
 [dev-dependencies]
 criterion = "0.2"
--- a/src/error/network_error.rs
+++ b/src/error/network_error.rs
@@ -25,6 +25,8 @@ pub enum NetworkErrorKind
     UDPSocketStateCreationFailed,
     #[fail(display = "The received data did not have any length.")]
     ReceivedDataToShort,
+    #[fail(display = "The protocol versions do not match.")]
+    ProtocolVersionMismatch,
 }
 
 #[derive(Debug)]
--- a/src/net/constants.rs
+++ b/src/net/constants.rs
@@ -1,7 +1,6 @@
-pub const FRAGMENT_HEADER_SIZE: u8 = 5;
-pub const PACKET_HEADER_SIZE: u8 = 10;
-pub const HEART_BEAT_HEADER_SIZE: u8 = 1;
-
+pub const FRAGMENT_HEADER_SIZE: u8 = 9;
+pub const PACKET_HEADER_SIZE: u8 = 14;
+pub const HEART_BEAT_HEADER_SIZE: u8 = 5;
 pub const MAX_FRAGMENTS_DEFAULT: u16 = 16;
 pub const FRAGMENT_SIZE_DEFAULT: u16 = 1024;
 /// Maximum transmission unit of the payload.
@@ -12,3 +11,9 @@ pub const FRAGMENT_SIZE_DEFAULT: u16 = 1024;
 /// This is not strictly guaranteed -- there may be less room in an ethernet frame than this due to
 /// variability in ipv6 header size.
 pub const DEFAULT_MTU: u16 = 1452;
+/// This is the current protocol version.
+///
+/// It is used for:
+/// - Generating crc32 for the packet header.
+/// - Validating if arriving packets have the same protocol version.
+pub const PROTOCOL_VERSION: &'static str = "laminar-0.1.0";
--- a/src/packet/packet_processor.rs
+++ b/src/packet/packet_processor.rs
@@ -5,8 +5,10 @@ use net::{SocketState, NetworkConfig};
 use packet::{header, Packet};
 use sequence_buffer::{SequenceBuffer, ReassemblyData};
 use self::header::{FragmentHeader, PacketHeader, HeaderReader};
-
-use error::{self, PacketErrorKind, FragmentErrorKind, NetworkResult};
+use protocol_version::ProtocolVersion;
+use super::PacketTypeId;
+use error::{self, PacketErrorKind, FragmentErrorKind, NetworkResult, NetworkErrorKind};
+use byteorder::{BigEndian, ReadBytesExt};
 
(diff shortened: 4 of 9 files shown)


Cargo.toml
```diff
@@ -29,6 +29,7 @@ failure = "0.1"
 failure_derive = "0.1"
 clap = { version = "2.32", features = ["yaml"], optional = true }
 env_logger = { version = "0.5.13", optional = true }
+crc = "1.8.1"
```
LucioFranco: should be `"1.8"`

src/error/network_error.rs
```diff
@@ -25,6 +25,8 @@ pub enum NetworkErrorKind
     UDPSocketStateCreationFailed,
     #[fail(display = "The received data did not have any length.")]
     ReceivedDataToShort,
+    #[fail(display = "The protocol version from the packet did not match the local protocol version.")]
```
LucioFranco: Please change this to `The protocol versions do not match`.

src/error/network_error.rs
```diff
@@ -25,6 +25,8 @@ pub enum NetworkErrorKind
...
+    ProtocolVersionMisMatch,
```
LucioFranco: I would change this to `ProtocolVersionMismatch`

src/net/constants.rs
```diff
@@ -11,4 +10,10 @@ pub const FRAGMENT_SIZE_DEFAULT: u16 = 1024;
...
 /// variability in ipv6 header size.
-pub const DEFAULT_MTU: u16 = 1452;
\ No newline at end of file
+pub const DEFAULT_MTU: u16 = 1452;
+/// This is the current protocol version.
+///
+/// It is used for:
+/// - Generating crc32 for the packet header.
+/// - Validating if arriving packets have the same protocol version.
+pub const PROTOCOL_VERSION: &'static str = "laminar-0.1.0";
```
LucioFranco: I want to think if there is a better way to do this. 🤔 . This is fine for now but if we do merge this, lets open an issue to follow up on.

src/packet/packet_processor.rs
```diff
@@ -28,14 +30,25 @@ impl PacketProcessor {
...
+            return Err(NetworkErrorKind::ProtocolVersionMisMatch.into());
+        }
+
+        // set cursor back to 0 because above values will be read again later in code.
+        cursor.set_position(0);
+
+        let received_bytes = match packet_id {
+            PacketTypeId::Packet => self.handle_normal_packet(&mut cursor, &addr, socket_state),
+            PacketTypeId::Fragment => self.handle_fragment(&mut cursor),
+            _ => {Ok(None) }
```
LucioFranco: Space here please after the first `{` 😄

src/packet/packet_processor.rs
```diff
@@ -222,7 +235,16 @@ mod tests {
...
     }
+
+    /// Tests if a packet will be processed right.
+///
+/// 1. first create test Packet
+/// 2. process it with `pre_process_packet` so we have valid raw data
+/// 3. then assert that the Packet we've gotten contains the right data.
+    #[test]
+    fn packets_ignored_with_different_protocol_id() {
+
```
LucioFranco: Lets add a test here please

src/packet/packet_processor.rs
```diff
@@ -222,7 +235,16 @@ mod tests {
                 is_packet_reassembled = true;
             }
         }
-
         assert!(is_packet_reassembled);
     }
+
+    /// Tests if a packet will be processed right.
+///
```
LucioFranco: Indent these in please.

src/protocol_version.rs
```diff
@@ -0,0 +1,37 @@
...
+{
+    pub fn get_version() -> &'static str
+    {
+        return PROTOCOL_VERSION
+    }
+
+    /// This will return the crc32 from the current protocol version.
+    pub fn get_crc32() -> u32 {
+        CALCULATE_CRC32.call_once(|| {
+            unsafe { VERSION_CRC32 = crc32::checksum_ieee(PROTOCOL_VERSION.as_bytes()); }
```
LucioFranco: So as I have learned if we call unsafe code and wrap it, we need to have some invariants that we check that guarantee that this is actually safe.

src/protocol_version.rs
```diff
@@ -0,0 +1,37 @@
...
+        });
+
+        unsafe { VERSION_CRC32 }
+    }
+
+    /// This will return the crc32 from the current protocol version.
+    pub fn valid_version(protocol_version_crc32: u32) -> bool {
+        protocol_version_crc32 == unsafe { VERSION_CRC32 }
```
LucioFranco: Same as above.

src/net/constants.rs
TimonPost: What is it that you don't like about it. You want somehow identify the current protocol I have seen this in a couple of libraries done like this. But if it is an issue lets discuss it in an issue.

src/protocol_version.rs
TimonPost: I am using `lazy_static` instead now. Think this is more cleaner.

src/net/constants.rs
LucioFranco: Its hard coded and I'd prefer to find a better place maybe to write this.

fhaynes: I think the idea would be that it gets bumped here if the protocol changes? I'm not sure where else you could put it, unless you wanted to break it up into multiple variables and assemble a String, but not sure how that would help.

TimonPost:
bors +r=fhaynes,LucioFranco

LucioFranco:
@TimonPost close, `bors r=fhaynes,LucioFranco` the `r+` is for when you are the only approver and reviewer.

src/net/constants.rs
LucioFranco: We could use a file called VERSION that is at the root and use `include_bytes!()`

TimonPost:
bors r=fhaynes,LucioFranco

PR #61, opened 2018-10-23
Remove a useless clone and fix a clippy warning.

CBenoit:
This is in `SocketState::pre_process_packet`.
A clippy lint was telling that packet wasn't consumed in the function. The cause is a call to `clone` on packet.
I first thought about taking a reference instead, but it doesn't reflect the correct API usage.
That's what I come up with. It removes a useless copy.

### DIFF
--- a/src/net/socket_state.rs
+++ b/src/net/socket_state.rs
@@ -44,7 +44,8 @@ impl SocketState {
         })
     }
 
-    /// This will initialize the seq number, ack number and give back the raw data of the packet with the updated information.
+    /// This will initialize the seq number, ack number and give back the raw data of the packet
+    /// with the updated information. As such the packet argument is consumed.
     pub fn pre_process_packet(
         &mut self,
         packet: Packet,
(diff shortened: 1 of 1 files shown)


src/net/socket_state.rs
```diff
@@ -120,20 +117,27 @@ impl SocketState {
...
+                    &packet.payload()[start_fragment_pos as usize..end_fragment_pos as usize]; /* upcast is safe */
 
                 packet_data.add_fragment(&fragment, fragment_data.to_vec());
             }
         }
 
-        let mut lock = connection
-            .write()
-            .map_err(|error| NetworkError::poisoned_connection_error(error.description()))?;
+        let packet_addr = packet.addr();
```
TimonPost: This line is only used once at line 140 so no extra allocation needed. Can get removed If you change line 140 to `packet.addr()`

src/net/socket_state.rs
```diff
@@ -120,20 +117,27 @@ impl SocketState {
...
+
+            // each time we send a packet we increase the local sequence number
+            lock.seq_num = lock.seq_num.wrapping_add(1);
+
+            // queue new packet
+            lock.waiting_packets.enqueue(connection_seq, packet);
+        }
 
-        Ok((packet.addr(), packet_data))
+        Ok((packet_addr, packet_data))
```
TimonPost: see above comment

src/net/socket_state.rs
CBenoit: It cannot be removed since packet is effectively consumed at line 137. You would have a use after move error.

PR #63, opened 2018-10-23
Implemented basic server-client

TimonPost:
Implemented basic server-client api.

You might want to review it but don't do it based on API it is relay some proof  of concept, I wrote all of it in just 2 hour's.

This branch needs to be merged on a separated branch. Because it has no direct relations to laminar. But we have decided to have it in the root folder. We later may move it to its own crate or integrate into laminar or amethyst_network

**disclaimer**
_Take note that this code is not working, it is only the API. Feel free to open an other PR to make the API better and spit your own code over it. All the function calls are just basic concepts and so look with an open vision to the code. And don't think every piece of code has an purpose full meaning._

### DIFF
--- a/laminar_abstractions/.gitignore
+++ b/laminar_abstractions/.gitignore
@@ -1,0 +1,2 @@
+/target
+**/*.rs.bk
--- a/laminar_abstractions/Cargo.toml
+++ b/laminar_abstractions/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "laminar-abstractions"
+version = "0.1.0"
+authors = ["Timon Post <timonpost@hotmail.nl>"]
+
+[dependencies.laminar]
+path = "../laminar"
+branch = "master"
+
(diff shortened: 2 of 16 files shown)


LucioFranco:
@TimonPost how does this fit into the RFC you just opened? 

Plus, I think we should find a better name for this library.

TimonPost:
Not sure what you mean? This is not related to the RFC I created. I agree we should find a better name had that already on my todo list :)

LucioFranco:
@TimonPost great, ill hold off on reviewing this, till after we have more time on the RFC 😄

LucioFranco:
This will need a rebase against `master`

LucioFranco:
@torkleyy removing this from `0.1.0` because this is experimental and we will revisit this once `0.1.0` is out.

torkleyy:
Yours to manage, so sure :wink: 

I was just guessing to have something to start with.

PR #64, opened 2018-10-24
First pass at a docker-compose based test with the testing binary.

fhaynes:


LucioFranco:
I kind want to put the scripts and compose file inside a folder, that way we can easily ignore them when we upload the package to crates.io

fhaynes:
> I kinda want to put the scripts and compose file inside a folder, that way we can easily ignore them when we upload the package to crates.io

The problem with that is that a Dockerfiles treats the folder it is in as its root so it couldn't reach out of that directory to get the binary. Pretty sure I can change that, though. There's a setting somewhere....maybe....

fhaynes:
And do you mean the Makefile too?

fhaynes:
Hmm, I don't see a reason to hide the Makefile, really. They're quite common...why would we hide it, but not, say, bors.toml?

fhaynes:
I did move the Dockerfile, though. They added a flag to let you specify which Dockerfile to use so it works fine in a sub-directory.

torkleyy:
A makefile in the directory can cause unnecessary rebuilds by Cargo AFAIK.

fhaynes:
@torkleyy that's interesting...do you have more details?

LucioFranco:
@fhaynes I don't need the makefile for building laminar? only for the tester so why would it be in the root? I would like to keep it clean and isolate tester scripts in their own folder.

fhaynes:
@LucioFranco I don't think a Makefile qualifies as dirty, given their commonality in this use case. And what do you mean by keeping it clean? You just don't want it on crates.io?

LucioFranco:
@fhaynes will the make file be used to build the crate for a library use? I see the makefie as only used for the testing suite, and thus its not required to compile the library. Therefore it should live outside the root.

fhaynes:
@LucioFranco that doesn't really answer either of my questions. Let me ask another way. What is the harm of it being in the root?

fhaynes:
And Makefiles are actually extremely convenient for building the library and binary, not just testing it. It just doesn't do that yet.

torkleyy:
@fhaynes Can't find it anymore. I think I could try it out though.

fhaynes:
@torkleyy Thanks, but after a long, long, _long_ discussion with @LucioFranco I'm going to move the Makefile. =)

torkleyy:
Sounds good, I'm in favour of moving, too ^^

src/bin/laminar-tester.rs
```diff
@@ -99,7 +101,15 @@ fn run_server(socket_addr: &str) {
 
 fn run_client(test_name: &str, destination: &str, endpoint: &str, pps: &str, test_duration: &str) {
     let network_config = net::NetworkConfig::default();
-    let mut client = net::UdpSocket::bind(endpoint, network_config.clone()).unwrap();
+    let mut client = match net::UdpSocket::bind(endpoint, network_config.clone()) {
+        Ok(c) => { c }
+        Err(e) => { 
+            println!("Error binding was: {:?}", e);
+            exit(1);
+
```
LucioFranco: extra space here 😄 , do you have rustfmt setup on save? if not I really recommend it.

fhaynes: I think we're covering that in our style guide discussion.

PR #68, opened 2018-10-28
WIP: Example Link Conditioner

fhaynes:
@TimonPost @LucioFranco this is a quick example of what I mean when I say link conditioner. The thing I don't like about this style is the extra if check per packet. Any ideas on how to avoid it?

### DIFF
--- a/src/net/link_conditioner.rs
+++ b/src/net/link_conditioner.rs
@@ -1,14 +1,14 @@
-//! This module provides means to simulate various network conditions for development. The primary focus is 
+//! This module provides means to simulate various network conditions for development. The primary focus is
 //! for testing applications under adverse conditions such as high packet loss networks, or high latency
 //! networks.
 
-use rand::prelude::*;
+use rand::prelude::random;
 
 pub struct LinkConditioner {
     // Value between 0 and 1, representing the % change a packet will be dropped on sending
     packet_loss: f64,
     // Value in milliseconds, representing the delay imposed between packets
-    latency: u32
+    latency: u32,
 }
 
 impl LinkConditioner {
@@ -16,7 +16,7 @@ impl LinkConditioner {
     pub fn new() -> LinkConditioner {
         LinkConditioner {
             packet_loss: 0.0,
-            latency: 0
+            latency: 0,
         }
     }
 
@@ -35,4 +35,3 @@ impl LinkConditioner {
         }
     }
 }
-
--- a/src/net/link_conditioner.rs
+++ b/src/net/link_conditioner.rs
@@ -1,0 +1,38 @@
+//! This module provides means to simulate various network conditions for development. The primary focus is 
+//! for testing applications under adverse conditions such as high packet loss networks, or high latency
+//! networks.
+
+use rand::prelude::*;
+
+pub struct LinkConditioner {
+    // Value between 0 and 1, representing the % change a packet will be dropped on sending
+    packet_loss: f64,
+    // Value in milliseconds, representing the delay imposed between packets
+    latency: u32
+}
+
+impl LinkConditioner {
+    /// Creates and returns a LinkConditioner
+    pub fn new() -> LinkConditioner {
+        LinkConditioner {
+            packet_loss: 0.0,
+            latency: 0
+        }
+    }
+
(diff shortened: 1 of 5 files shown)


fhaynes:
I thought we were against that? =P

LucioFranco:
@fhaynes no 🤔 , I think this would fit under the `tester` feature?

fhaynes:
I'm going to just...just...hug you to death when I see you.

src/net/udp.rs
```diff
@@ -51,14 +54,30 @@ impl UdpSocket {
         let (addr, mut packet_data) = self.state.pre_process_packet(packet, &self.config)?;
 
         let mut bytes_sent = 0;
-
-        for payload in packet_data.parts() {
-            bytes_sent += self
-                .socket
-                .send_to(&payload, addr)
-                .map_err(|io| NetworkError::from(NetworkErrorKind::IOError { inner: io }))?;
+        if cfg!(feature = "tester") {
```
LucioFranco: Don't think this does the selective compilation? This will check at runtime?

src/net/link_conditioner.rs
```diff
@@ -0,0 +1,38 @@
+//! This module provides means to simulate various network conditions for development. The primary focus is 
+//! for testing applications under adverse conditions such as high packet loss networks, or high latency
+//! networks.
+
+use rand::prelude::*;
```
LucioFranco: Disclamer: this is my personal opinion on this

but I would rather not use `*`, just explicitly import the function if you can.

src/net/udp.rs
fhaynes: https://doc.rust-lang.org/book/first-edition/conditional-compilation.html

Scroll to bottom please and tell me it doesn't mean what I think it means

LucioFranco: Okay, so it still compiles the code just makes the branch unreachable whereas with the attribute it doesn't even compile that branch?

src/net/link_conditioner.rs
fhaynes: omg working on it. Like, I just wrote this 10 minutes ago.

LucioFranco: There is a reason they call me the fastest review in the west 😉

fhaynes: YOU LIVE ON THE EAST COAST

LucioFranco:
This is looking good! I like the concept.

src/net/udp.rs
fhaynes: Yes, I think so? What I am wanting to avoid is having an entirely new module

LucioFranco: Right, I know this might be a bit messy. What I want to avoid is exposing testing interfaces to the user as well.

but now that I think about it, it might be useful to say, flip a switch in amethyst and simulate poor link quality. 🤔 

Maybe we can make it a trait and have different types.

fhaynes: The trait idea had occurred to me, and people have expressed interest in having it available during runtime. Honestly the if check is going to add latency on the order of `ns` is that. For 0.1.0, we could just leave it in-line and use benchmarks to see if it causes problems?

LucioFranco: Are you sure, if there is a hard coded pass through, like say if `None` is used would LLVM optimize it out? aka remove the unreachable branch?

fhaynes: I don't know about LLVM optimizing it out, but a boolean if check is very, very fast.

fhaynes:
OK, I'm going to propose a radical idea. We leave it in-line in the production code for 0.1.0, and we'll use our benchmarks to determine if we should go with a trait or conditional compilation in 0.2.0. A boolean if check is not going to have a perceptible impact on performance, and if we wanted, there's tons of optimizations we could do that would make up for it.

fhaynes:
Closing this in favor of doing a proper one from my fork

LucioFranco:
@fhaynes shall we delete the stale branch then?

fhaynes:
> @fhaynes shall we delete the stale branch then?

Yeah, working on it! Somebody had their cheerios this morning

PR #73, opened 2018-10-28
Add simple link conditioner

fhaynes:
I think we should leave this inline for 0.1. A boolean if check is going to add a minuscule amount of latency, and we can revisit conditional compilation or a trait in 0.2.0.


fhaynes:
omg, closing this one, cause I dunno wtf happened with the commits. will redo.

PR #86, opened 2018-11-07
Implemented re sending of packets.

TimonPost:
I have created a function called: `gather_dropped_packets()` this will gather dropped packets by draining them from all reliable channels. This function is called in `udp.rs` `send()` so that before the user his packet goes out there will be a check if there are dropped packets that needs to be resend, if there are those will be resend until they are being aknowleged.

**A question I have:**

I don't know how to test whether this really works. Like our acknowledgment system has unit-test for it. But I can't specify check if dropped packets are resent using our UDP socket. @fhaynes is there a way this could be covered by your scenario tests?

### DIFF
--- a/src/infrastructure/channels/reliable_channel.rs
+++ b/src/infrastructure/channels/reliable_channel.rs
@@ -60,6 +60,30 @@ impl ReliableChannel {
             rtt: 0.0,
         }
     }
+
+    /// Check if this channel has dropped packets.
+    ///
+    /// You could directly call `ReliableChannel::drain_dropped_packets()` and if it returns an empty vector you know there are no packets.
+    /// But draining a vector will have its extra check logic even if it's empty.
+    /// So that's why this function exists just a little shortcut to check if there are dropped packets which will be faster at the end.
+    pub fn has_dropped_packets(&self) -> bool {
+        !self.dropped_packets.is_empty()
+    }
+
+    /// Creates a draining iterator that removes dropped packets and yield the ones that are removed.
+    ///
+    /// So why drain?
+    /// You have to think about the packet flow first.
+    /// 1. Once we send a packet we place it in a queue until acknowledged.
+    /// 2. If the packet doesn't get acknowledged in some time it will be dropped.
+    /// 3. When we notice the packet drop we directly want to resend the dropped packet.
+    /// 4. Once we notice that we start at '1' again.
+    ///
+    /// So keeping track of old dropped packets does not make sense, at least for now.
+    /// We except when dropped packets are retrieved they will be sent out so we don't need to keep track of them internally the caller of this function will have ownership over them after the call.
+    pub fn drain_dropped_packets(&mut self) -> Vec<Box<[u8]>> {
+        return self.dropped_packets.drain(..).collect()
+    }
 }
 
 impl Channel for ReliableChannel {
@@ -141,9 +165,6 @@ impl Channel for ReliableChannel {
 
         self.dropped_packets = dropped_packets.into_iter().map(|(_, p)| p).collect();
 
-        let a = buffer.len();
-        let b = acked_header.size();
-        // TODO: resent packets if there are dropped packets
         Ok(&buffer[acked_header.size() as usize .. buffer.len()])
     }
 }
--- a/src/net/connection/virtual_connection.rs
+++ b/src/net/connection/virtual_connection.rs
@@ -109,6 +109,17 @@ impl VirtualConnection {
 
         Ok(Some(Packet::new(self.remote_address, Box::from(payload), header.delivery_method)))
     }
+
+    /// This will gather dropped packets from the reliable channels.
+    ///
+    /// Note that after requesting dropped packets the dropped packets will be removed from this client.
+    pub fn gather_dropped_packets(&mut self) -> Vec<Box<[u8]>> {
+        if self.reliable_unordered_channel.has_dropped_packets() {
+            return self.reliable_unordered_channel.drain_dropped_packets();
+        }
+
+        Vec::new()
+    }
 }
 
 impl fmt::Debug for VirtualConnection {
@@ -127,7 +138,7 @@ mod tests {
     use net::NetworkConfig;
     use net::connection::VirtualConnection;
     use packet::header::{AckedPacketHeader, StandardHeader, HeaderReader, HeaderParser};
-    use infrastructure::DeliveryMethod;
+    use infrastructure::{DeliveryMethod, ReliableChannel, Channel};
     use packet::PacketTypeId;
     use net::constants::STANDARD_HEADER_SIZE;
     use std::sync::Arc;
--- a/src/net/udp.rs
+++ b/src/net/udp.rs
@@ -1,4 +1,4 @@
-use std::net::{self, ToSocketAddrs};
+use std::net::{self, ToSocketAddrs, SocketAddr};
 use net::connection::ConnectionPool;
 
 use error::{NetworkError, NetworkErrorKind, NetworkResult};
@@ -80,22 +80,33 @@ impl UdpSocket {
         if let Some(link_conditioner) = &self.link_conditioner {
             if link_conditioner.should_send() {
                 for payload in packet_data.parts() {
-                    bytes_sent += self.socket.send_to(&payload, packet.addr()).map_err(|io| {
-                        NetworkError::from(NetworkErrorKind::IOError { inner: io })
-                    })?;
+                    bytes_sent += self.send_packet(&packet.addr(), &payload)?;
                 }
             }
         } else {
+            for payload in lock.gather_dropped_packets() {
+                bytes_sent += self.send_packet(&packet.addr(), &payload)?;
+            }
+
             for payload in packet_data.parts() {
-                bytes_sent += self
-                .socket
-                .send_to(payload, packet.addr())
-                .map_err( |io | NetworkError::from(NetworkErrorKind::IOError { inner: io })) ?;
+                bytes_sent += self.send_packet(&packet.addr(), &payload)?;
             }
         }
 
-    Ok(bytes_sent)
-}
+        Ok(bytes_sent)
+    }
+
+    /// Send a single packet over the udp socket.
+    fn send_packet(&self, addr: &SocketAddr, payload: &[u8]) -> NetworkResult<usize>  {
+        let mut bytes_sent = 0;
+
+        bytes_sent += self
+            .socket
+            .send_to(payload, addr)
+            .map_err(|io| NetworkError::from(NetworkErrorKind::IOError { inner: io }))?;
+
+        Ok(bytes_sent)
+    }
 
     /// Sets the blocking mode of the socket. In non-blocking mode, recv_from will not block if there is no data to be read. In blocking mode, it will. If using non-blocking mode, it is important to wait some amount of time between iterations, or it will quickly use all CPU available
     pub fn set_nonblocking(&mut self, nonblocking: bool) -> NetworkResult<()> {


src/infrastructure/channels/reliable_channel.rs
```diff
@@ -141,9 +141,19 @@ impl Channel for ReliableChannel {
...
-        // TODO: resent packets if there are dropped packets
         Ok(&buffer[acked_header.size() as usize .. buffer.len()])
     }
+}
+
+impl ReliableChannel {
+    /// Check if this channel has dropped packets.
+    /// Performance wise it is better to use this to check if there are dropped packets than calling 'get_dropped_packets directly()'.
+    pub fn has_dropped_packets(&self) -> bool {
+        !self.dropped_packets.len() == 0
```
LucioFranco: please use `is_empty` here

src/infrastructure/channels/reliable_channel.rs
```diff
@@ -141,9 +141,19 @@ impl Channel for ReliableChannel {
...
+    }
+
+    /// Creates a draining iterator that removes dropped packets and yield the ones that are removed.
+    pub fn drain_dropped_packets(&mut self) -> Vec<Box<[u8]>> {
+        return self.dropped_packets.drain(..).collect()
```
LucioFranco: could you explain why you use drain here?

src/net/connection/virtual_connection.rs
```diff
@@ -109,6 +109,17 @@ impl VirtualConnection {
 
         Ok(Some(Packet::new(self.remote_address, Box::from(payload), header.delivery_method)))
     }
+
+    /// This will gather dropped packets from the reliable channels.
+    ///
+    /// Note that after requesting dropped packets the dropped packets will be removed from this client.
+    pub fn gather_dropped_packets(&mut self) -> Option<Vec<Box<[u8]>>>{
```
LucioFranco: This could just be an empty Vec btw

src/net/connection/virtual_connection.rs
```diff
@@ -109,6 +109,17 @@ impl VirtualConnection {
...
+        if self.reliable_unordered_channel.has_dropped_packets() {
+            return Some(self.reliable_unordered_channel.drain_dropped_packets());
```
LucioFranco: no returns please 😄

src/net/udp.rs
```diff
@@ -80,22 +80,35 @@ impl UdpSocket {
...
                 for payload in packet_data.parts() {
-                    bytes_sent += self.socket.send_to(&payload, packet.addr()).map_err(|io| {
-                        NetworkError::from(NetworkErrorKind::IOError { inner: io })
-                    })?;
+                    bytes_sent += self.send_packet(&packet.addr(), payload)?;
                 }
             }
         } else {
+            if let Some(dropped) = lock.gather_dropped_packets() {
+                for payload in dropped {
```
LucioFranco: as I said about you just need to create an `iter` here for dropped packets. if the iter is empty it doesn't run the code within the loop.

src/net/udp.rs
```diff
@@ -80,22 +80,35 @@ impl UdpSocket {
...
+    /// Send a single packet over the udp socket.
+    fn send_packet(&self, addr: &SocketAddr, payload: &[u8]) -> NetworkResult<usize>  {
+        let mut bytes_sent = 0;
+
+        bytes_sent += self
+            .socket
+            .send_to(payload, addr)
+            .map_err(|io| NetworkError::from(NetworkErrorKind::IOError { inner: io }))?;
+
+        return Ok(bytes_sent)
```
LucioFranco: drop the return 😄

src/infrastructure/channels/reliable_channel.rs
```diff
@@ -141,9 +141,19 @@ impl Channel for ReliableChannel {
...
         self.dropped_packets = dropped_packets.into_iter().map(|(_, p)| p).collect();
 
-        let a = buffer.len();
-        let b = acked_header.size();
-        // TODO: resent packets if there are dropped packets
         Ok(&buffer[acked_header.size() as usize .. buffer.len()])
     }
+}
+
+impl ReliableChannel {
```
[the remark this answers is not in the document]
fhaynes: If this is something we want to do consistently, please put it in the style guide.

src/infrastructure/channels/reliable_channel.rs
fhaynes: Please make sure to explain why when making these comments.

src/infrastructure/channels/reliable_channel.rs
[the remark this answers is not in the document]
LucioFranco: More of a suggestion

src/infrastructure/channels/reliable_channel.rs
fhaynes: What is the specific question you are asking here? Is there a reason to not use drain?

LucioFranco: Yes, why is drain used here when you can just return the vector?

fhaynes: @TimonPost The explanation for this is something that we should probably add to the docstring.

src/net/connection/virtual_connection.rs
fhaynes: Clippy lint should be catching these, and if we aren't running it is as part of the CI/CD, we should

src/net/udp.rs
fhaynes: Clippy lint should be catching these, and if we aren't running it is as part of the CI/CD, we should

LucioFranco:
As for your question, I think you can make an integration test that drops the packet possibly? You will have to mock out some stuff though.

src/infrastructure/channels/reliable_channel.rs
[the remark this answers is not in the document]
fhaynes: For a code review, let's keep one-off style suggestions to some other venue so they don't clutter things.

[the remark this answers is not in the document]
fhaynes: Or at the very least, mark them as a suggestion explicitly in the PR, and then if the person doesn't follow it, ignore it.

fhaynes:
> As for your question, I think you can make an integration test that drops the packet possibly? You will have to mock out some stuff though.

I did write that whole link conditioner thing....

src/infrastructure/channels/reliable_channel.rs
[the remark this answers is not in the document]
LucioFranco: Well, I find it easier to get that change in here and do multiple passes of the PR. If we decide that it's not needed etc the comment can be resolved manually.

if we do the change after there is a really high chance it will be forgotten.

LucioFranco:
@fhaynes yes, @TimonPost use that, long day am tired. 😴

src/net/connection/virtual_connection.rs
fhaynes: That's a personal style thing, lots of people prefer returns. However, no returns is as close to an axiom as you'll find in Rust style I think, so I'm fine with keeping them out. Let's just automate it.

src/infrastructure/channels/reliable_channel.rs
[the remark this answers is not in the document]
fhaynes: So it isn't more of a suggestion? =)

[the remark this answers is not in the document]
LucioFranco: No, I presented my reason, if he has a reason that is better than we can ignore it. Its more of to start a conversation about it, instead of leaving it there. You are not required to fix every issue requested in a `request changes` review.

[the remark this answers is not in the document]
fhaynes: Saying "get that change in here" implies that it isn't just a suggestion, it is a change you want to see happen. So, style guide. If it is just a suggestion along the lines of "I find it easier if I do...", once is enough. The person will like it and follow it, or not.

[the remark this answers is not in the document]
fhaynes: We have a lot of comments in our PRs that turn into long style discussions, and yet we haven't codified and enforced any that I know of. Either we don't care enough to do the work to do this properly and we can stop making comments about them and accept them, or do we do care enough to do the work to ensure a common set of rules and automation.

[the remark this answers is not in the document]
TimonPost: I would say if that's the still we are maintaining let's do that. I'll update the code and update also our style guide dock.

src/infrastructure/channels/reliable_channel.rs
TimonPost: Oké, I'll have some more detailed comment. Also why use 'is empty', question I than have is what is empty? I'll see what I can do for a better comment.

src/infrastructure/channels/reliable_channel.rs
TimonPost: I use drain because otherwise dropped packets will be queued up in the vector. Once you request dropped packets you are supposed to resend them if they will not be acknowledged thw packets will aromatically end in the dropped vector. If where were not to drain the collection we can't keep proper track of dropped packets.

I'll update the docstring

src/net/connection/virtual_connection.rs
TimonPost: Reason why I do that is because I don't have to allocate a new vector if there are no dropped packets. I could simply return None, isn't this a better approach?

src/net/connection/virtual_connection.rs
TimonPost: I am actually prefering return is most cases. But in this scenario I ended up with a compiler error. I'll revisit it and see what I can do.

src/net/udp.rs
TimonPost: Alright 😐

src/net/connection/virtual_connection.rs
fhaynes: I believe you can't in this case because of the if structure.

src/infrastructure/channels/reliable_channel.rs
fhaynes: `len() == 0` is a common enough pattern I don't see it as unclear. Also not all things implement is_empty().

LucioFranco: Its about reducing thought when its comes to reading a line having a `!somecode == 0` takes a lot more thought than `!somecode.is_empty()`.

fhaynes: Why do you think it takes a lot more thought?

(Yes, I am challenging a base assumption)

src/net/connection/virtual_connection.rs
LucioFranco: Creating an empty `Vec` doesn't allocate anything.

fhaynes: Nitpick: with_capacity is an option. =)

fhaynes: And it does still take up a minuscule amount of bookkeeping memory

src/infrastructure/channels/reliable_channel.rs
LucioFranco: because `is_empty` is explicit, its reads like english. `len() == 0` takes slightly more time to think about when reading it I find.I don't see why you would pick `len() == 0` over `is_empty`?

src/net/connection/virtual_connection.rs
LucioFranco: @TimonPost if you want we can look at it together in a play.rust-lang.org

src/infrastructure/channels/reliable_channel.rs
fhaynes: `len() == 0` is just as explicit. You're right in that is_empty() is much English-ish. Should that be a primary concern when many of our devs speak English as a 2nd language?

LucioFranco: I mean, the codebase is in english so therefore it should try and flow as English as possible?

fhaynes: Excellence does not require perfection.
Perfection is the enemy of progress.

TimonPost: I have implemented `is_empty()` instead of manually checking the length of the vector. I did not know of the function `is_empty()` but it basically does the same as you look at stdlib.

src/net/connection/virtual_connection.rs
TimonPost: @fhaynes should I go with Vec::with_capacity(0) or just Option. I have been at that decision and went with Option but watch your view?

src/net/connection/virtual_connection.rs
TimonPost: There you are: https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=faaec323b3e789cd878cf1092fe2b4a8

src/net/udp.rs
TimonPost: Done

src/net/connection/virtual_connection.rs
fhaynes: Calling this one resolved. Make an issue for it if someone feels it needs to be revisited.

src/net/connection/virtual_connection.rs
fhaynes: Talked with @TimonPost on Discord about this. Summary was to return `Vec::with_capacity(0)`.

Issue #70, opened 2018-10-28
Transition to use macro instead of macro_use


LucioFranco:
Now that rust `1.30` is stable and out. We can use `use macro` instead of the ambiguous `#[macro_use]`.

This issue should be easy and I am opening it up for a good first issue. If anyone needs more direction please ask in here or on discord 😄

TimonPost:
This can be closed now right?

PR #91, opened 2018-11-09
Replace #[macro_use] by use macro.

CBenoit:
See https://github.com/amethyst/laminar/issues/70


LucioFranco:
Actually going to have to wait for the new beta build to propagate.

Issue #82, opened 2018-11-02
Restart connection thread when panicked.


TimonPost:
Currently, our connection thread can panic put it will not be restarted. We need to find a way to restart this thread.

PR #99, opened 2018-11-11
Changed some panics to errors so the timeout checking thread won't fail

fhaynes:
Looks like @TimonPost already did most of it.


fhaynes:
Resolves #82

TimonPost:
How did this resolve the issue? You'd only removed panics.

LucioFranco:
Yeah as @TimonPost says, this will just continue to spew errors because the lock will continue to be in the Poisoned state.

fhaynes:
Closing in favor of https://github.com/amethyst/laminar/pull/101 which solves this and more in a better way.

PR #100, opened 2018-11-11
Ran rustfmt with all the defaults set to Edition 2018. 

fhaynes:
You'll need to use `cargo +nightly fmt` to run this locally. These are the default settings, so if ya'll don't like the look of some of it, we can take it out. The options are in `.rustfmt.toml` in the root.


fhaynes:
Closing this in favor of using stable

Issue #67, opened 2018-10-26
Re-work client counting to return a result


fhaynes:
```rust
    pub fn count(&self) -> usize {
         match self.connections.read() {
             Ok(connections) => { connections.len() },
             Err(_) => { 0 },
```
Should probably return a result, not swallow the error and return 0.

TimonPost:
One thing I don't like about this have you ever called a `count()` function which returns a result? I didn't. I find it a weird idea that something simple like `count()` can go wrong. So an alternative would be to add a counter to the connection pool who is counting the connections. This way we don't have to acquire the read lock at all.

fhaynes:
Well, my primary concern is that it returns 0 if there is an error, which could also be a valid number of connections. In C-land, I'd suggest returning -1. How about Option<usize>?

TimonPost:
Indeed you are right with it that there could be `0` connections. I like `Option` more. But why not have a counter for the connections. Who will increase when incoming connection and decreases when connections times out. Then if the user wants to request the connected clients it could never fail and there is no lock needed.

fhaynes:
Even better! Sorry, hadn't had coffee and missed that part earlier.

LucioFranco:
@TimonPost yes, I think that is a good solution 👍

TimonPost:
I've done some things but since clients is added from within the timeout thread we can not access the counter from `ConnectionPool` other than using an `Arc` with `Mutex`. But since this will also be error-prone when getting the count we need a redisgn. 

We should move the connection thread out of `ConnectionPool`. Because I want to have `ConnectionPool` to be one type with one responsibility. 

Connection pool should have the following methods:

```
get_connection_or_insert();
remove_by_address();
count();
```

Next, the timeout loop currently inside the connection pool should be placed somewhere else. This way we can handle the client count. Like when function `remove_by_address()` is called we decrement the client count and when `get_connection_or_insert()` we increment the client count. The connection manager has ownership of its internals. And the timeout thread is not responsible for updating the client count.

TimonPost:
#101 fixes it

PR #101, opened 2018-11-11
67 connection count fix

TimonPost:
### This pr contains: 

**_1. Some refactor._**
We now have the `ConnectionPool` to be a type on its own. And we have the `TimeoutThread` doing its job and having a reference to the connection pool. This way separate concerns. Manly this design decision came up when worked on the count problem. 
**ConnectionPool:**
This will only be responsible for managing things related to the connection collection. You can think of it as a vector with only the `remove_connection`, `count`, `get_connection_or_insert` functions. This also has the function to return connections timed out. 
**TimoutThread:**
This will be responsible for things like thread management, the join handle or example. Also, we can do things like restarting, starting, stopping the thread in here.

**_2. Fixed connection count #67._**
I did go for `Result` instead of the count variable we spoke of in #67. I did this for some valid reasons. 
I can explain it if you really want it. But there are a lot of things that play a role. Just in short, because we need to mutate the count variable in `ConnectionPool` we need to make the functions doing that having a mutable self. If we do this the `timout_thread` needs to have access to a mutable `ConnectionPool`. This would lead up putting connection pool inside a `Mutex`.  If we do this we have to acquire the lock each time on packet receive and send. 

I don't think that is optimal. And for the use-cases of the `count` function, I think it is a bad idea. `count` is only used in our tests for now. So the best easiest fix is just let it return a `Result` of `usize`. 

But as I said there are more options to solve this problem wich will be to much to explain them all here. I had some text describing all the things playing a role here, but I lost it. Just trust me that for now this is te best option. If don't agree feel free to try it your self onces this get's merged.


src/net/connection/connection_pool.rs
```diff
@@ -1,53 +1,31 @@
...
         }
     }
 
-    /// Insert connection if it does not exists.
-    pub fn get_connection_or_insert(&mut self, addr: &SocketAddr) -> NetworkResult<Connection> {
+    /// Try getting connection by address if the connection does not exists it will be inserted.
+    pub fn get_connection_or_insert(&self, addr: &SocketAddr) -> NetworkResult<Connection> {
         let mut lock = self
             .connections
             .write()
```
LucioFranco: I thought we fixed this? If the connection exists we will acquire a write lock but we should only be acquiring the read lock?

src/net/connection/timeout_thread.rs
```diff
@@ -0,0 +1,82 @@
...
+        let sender = self.sender.clone();
+        let (tx, rx) = channel();
+
+        let thread = thread::spawn(move ||
+            loop {
+                match connection_pool.check_for_timeouts(poll_interval, &sender) {
+                    Ok(timed_out_clients) => {
+                        for timed_out_client in timed_out_clients {
+                            if let Err(e) = connection_pool.remove_connection(&timed_out_client) {
+                                let _ = tx.send(e);
```
LucioFranco: btw this just throws away the Result? Should we not handle this?

src/net/connection/connection_pool.rs
TimonPost: Might me wrong but for inserting we need write

src/net/connection/timeout_thread.rs
TimonPost: Mmm I think we are sure the receiving side will always exists. So there would not be any error. Send could only fail if receiver side does not exists if I am right.

src/net/connection/connection_pool.rs
fhaynes: Inserting a connection requires a write lock, yes

src/net/connection/timeout_thread.rs
LucioFranco: Then can this be an `expect`?

src/net/connection/connection_pool.rs
LucioFranco: fine we can address this later.

fhaynes: You could get a read lock and check if it exists, and then do a write lock only if an insert is needed. But we can make an issue for that.

fhaynes: #102

src/net/connection/timeout_thread.rs
fhaynes: With the mpsc channels, it would only ever fail if the receiving side does not exist. The point of the error is that it contains the data so it can be recovered if desired.

fhaynes:
bors r=LucioFranco, TimonPost

PR #103, opened 2018-11-11
WIP | Sequenced

TimonPost:
This is just to show of about how packets could be processed sequenced. We don't have to merge it into 0.1.0 however it is working and tested.

### DIFF
--- a/src/infrastructure/channels/mod.rs
+++ b/src/infrastructure/channels/mod.rs
@@ -22,5 +22,5 @@ pub trait Channel {
     ) -> NetworkResult<PacketData>;
 
     /// Progress an packet on receive and receive the processed data.
-    fn process_incoming<'d>(&mut self, buffer: &'d [u8]) -> NetworkResult<&'d [u8]>;
+    fn process_incoming<'d>(&mut self, buffer: &'d [u8]) -> NetworkResult<Option<&'d [u8]>>;
 }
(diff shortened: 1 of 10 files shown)


src/infrastructure/channels/mod.rs
```diff
@@ -22,5 +22,5 @@ pub trait Channel {
     ) -> NetworkResult<PacketData>;
 
     /// Progress an packet on receive and receive the processed data.
```
LucioFranco: Since you changed this to option can you add a doc comment that explains when an option value will be `Some` or `None`.

src/packet/packet_structure.rs
```diff
@@ -57,6 +57,24 @@ impl Packet {
...
 
+    /// Unreliable. Packet can be dropped, but never duplicated and arrive in order.
+    ///
+    /// *Details*
+    ///
+    /// |   Packet Drop   | Packet Duplication | Packet Order     | Packet Fragmentation | Packet Delivery |
+    /// | :-------------: | :-------------:    | :-------------:  | :-------------:      | :-------------: |
+    /// |       Yes       |      No            |      Yes         |      No             |       No        |
+    ///
+    /// Sequence means toss away any packets that are older than the most recent (like a position update, you don't care about older ones),
```
Abendstolz: Nitpicking: shouldn't this say "Sequence**d** means toss away..."?

PR #113, opened 2018-11-13
Fix code style via rustfmt

LucioFranco:
This formats our codebase with rustfmt:

```bash
$ cargo fmt --version
rustfmt 0.99.4-stable (1c408818 2018-08-27)
```


LucioFranco:
Next, step will be to enforce this version of rustfmt.

bors r+

Issue #115, opened 2018-11-19
Try to make 'recv()' funciton to be a imutable self.


TimonPost:
Currently, our UDP API has the `send` and `recv` functions who are taking in a mutable self. I have found a way to remove `mut` from `send()` but couldn't just remove mut from `recv()`

So why is this important? in 'amethyst_netork' I wanted to try separating sending and receiving by moving them to their own threads. I wanted to do this because there was some packet drop.  However, if you want to access the UDP socket in both threads you have to put this socket into an `Arc`. But now we get to the point where `recv()` is a mutable function, so we need a mutable reference to `UdpSocket`. We could then move the `UdpSocket` into its own `Mutex`. 

This has some problems:
1. When you want to receive with blocking on the lock stays unlocked during the duration of the receiving operation.
2. When you are receiving with nonblocking on you need to acquire the lock often.
  Mostly when using nonblocking you are doing inside a loop.  So each loop you have to acquire the lock which doesn't sound good to me. 

The best case scenario we need to make the `recv` function taking in immutable `self`, also the STDLIB `UdpSocket::recv()` is also immutable. They accomplished it by letting the user pass in a mutable buffer to write to. But this will be a little difficult with our current design since the `recv` output is an `Option` which could be `None`. 

Any thoughts on this how we could make this happen?

_our UDP API_
```
    pub fn recv(&mut self) -> NetworkResult<Option<Packet>> {
        let (len, addr) = self.socket.recv_from(&mut self.recv_buffer)?;

```

riemass:
After reading the discussion in the closed pull requests, the downside for the above solution is the extra time needed for the additional Vec allocation. There is also the option of putting the receive buffer on the stack:
```
let mut recv_buffer: [u8; DEFAULT_MTU as usize] = [0; DEFAULT_MTU as usize];
let (len, addr) = self.socket.recv_from(&mut recv_buffer[0..self._config.receive_buffer_max_size])?;
```
The downside is that the stack size would grow by 1500 bytes even when the receive_buffer_max_size is set to a lower value by the user.

TimonPost:
However, it still has to create a buffer on the stack for each time a packet is received. This is not optimal and we need to avoid it, if possible.

LucioFranco:
This should be a non-problem with mio since read and writes won't block and laminar will not have to create any threads.

TimonPost:
Fixed with #138

PR #116, opened 2018-11-19
Removed mut from the UdpSocket API.

TimonPost:
#115

### DIFF
--- a/src/net/udp.rs
+++ b/src/net/udp.rs
@@ -14,7 +14,6 @@ use std::sync::Arc;
 /// Represents an <ip>:<port> combination listening for UDP traffic
 pub struct UdpSocket {
     socket: net::UdpSocket,
-    recv_buffer: Vec<u8>,
     _config: Arc<NetworkConfig>,
     link_conditioner: Option<LinkConditioner>,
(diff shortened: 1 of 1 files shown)


src/net/udp.rs
```diff
@@ -47,11 +45,12 @@ impl UdpSocket {
     }
 
     /// Receives a single datagram message on the socket. On success, returns the packet containing origin and data.
-    pub fn recv(&mut self) -> NetworkResult<Option<Packet>> {
-        let (len, addr) = self.socket.recv_from(&mut self.recv_buffer)?;
+    pub fn recv(&self) -> NetworkResult<Option<Packet>> {
+        let mut recv_buffer = vec![0; self._config.receive_buffer_max_size];
```
LucioFranco: so this means everytime we call `recv` we allocate a new buffer? Doesn't sound very efficient.

TimonPost: True if you have alternatives please let me know under the linked issue.

TimonPost: I'm not sure on how to solve this problem right way other then doing this

LucioFranco:
I am going to close this as this is not the correct way to go about this.

Issue #102, opened 2018-11-11
Enhance locking when getting or inserting connections


fhaynes:
In `connection_pool.rs`, we have this function:
```rust
    pub fn get_connection_or_insert(&self, addr: &SocketAddr) -> NetworkResult<Connection> {
        let mut lock = self
            .connections
            .write()
            .map_err(|error| NetworkError::poisoned_connection_error(error.description()))?;

        let connection = lock
            .entry(*addr)
            .or_insert_with(|| Arc::new(RwLock::new(VirtualConnection::new(*addr, &self.config))));

        Ok(connection.clone())
    }
```
It acquires a write lock. It could be re-structured so that a read lock is acquired to check if an insertion is needed, and if so, acquire a write lock. A minor optimization.

TimonPost:
Solved in #119

PR #119, opened 2018-12-01
Optimized locking connection pool

TimonPost:
#102

### DIFF
--- a/src/net/connection/connection_pool.rs
+++ b/src/net/connection/connection_pool.rs
@@ -26,16 +26,32 @@ impl ConnectionPool {
 
     /// Try getting connection by address if the connection does not exists it will be inserted.
     pub fn get_connection_or_insert(&self, addr: &SocketAddr) -> NetworkResult<Connection> {
-        let mut lock = self
+        let lock = self
             .connections
-            .write()
+            .read()
             .map_err(|error| NetworkError::poisoned_connection_error(error.description()))?;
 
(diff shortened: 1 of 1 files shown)


src/net/connection/connection_pool.rs
```diff
@@ -26,16 +26,23 @@ impl ConnectionPool {
...
-            .write()
+        let lock =  self.connections
+            .read()
             .map_err(|error| NetworkError::poisoned_connection_error(error.description()))?;
 
-        let connection = lock
-            .entry(*addr)
-            .or_insert_with(|| Arc::new(RwLock::new(VirtualConnection::new(*addr, &self.config))));
+        if lock.contains_key(addr) {
+            return Ok(lock.get(addr).unwrap().clone());
```
LucioFranco: Can you remove the return and the unwrap?

src/net/connection/connection_pool.rs
```diff
@@ -26,16 +26,23 @@ impl ConnectionPool {
...
+        }else {
+            drop(lock);
```
LucioFranco: Good, so we don't create a deadlock 👍

src/net/connection/connection_pool.rs
TimonPost: Strange though I already did that

TimonPost:
bors r=LucioFranco

PR #121, opened 2018-12-02
Remove the mutex around VERSION_CRC32

jstnlef:
I suspect locking and unwrapping the mutex is a bit slower than just getting a copy of the u32.


TimonPost:
bors r=TimonPost

PR #123, opened 2018-12-02
Fix #122 - TimeoutThread: Kill system thread when dropped

### LINKED ISSUE #122: TimeoutThread does not stop when UdPSocket gets disposed.
When the UdpSocket goes out of scope the TimeoutThread will not be aborted. And there is no way to close it.

A possible solution is: https://github.com/Philpax/laminar/commit/165d4788a9d53ad387b4bcc89c45e436b437c5b3

Philpax:
This PR fixes the issue described in #122 where the underlying thread for `TimeoutThread` is not shut down upon termination of `TimeoutThread`.

This is done by implementing `Drop` for `TimeoutThread`, and adding `TimeoutThread` to `UdpSocket` so that it is dropped when the socket is dropped.

### DIFF
--- a/src/net/connection/timeout_thread.rs
+++ b/src/net/connection/timeout_thread.rs
@@ -1,7 +1,9 @@
 use super::ConnectionPool;
 use error::{NetworkError, NetworkErrorKind, NetworkResult};
 use events::Event;
+use log::error;
 
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::mpsc::{channel, Receiver, Sender};
 use std::sync::Arc;
 use std::thread;
@@ -19,6 +21,7 @@ pub const TIMEOUT_POLL_INTERVAL: u64 = 1;
 /// 3. Check if the last time we have heard from them (received a packet from them) is greater than the amount of time considered to be a timeout.
 /// 4. If they have timed out, send a notification up the stack.
 pub struct TimeoutThread {
+    shutdown_signal: Arc<AtomicBool>,
     poll_interval: Duration,
     timeout_check_thread: Option<thread::JoinHandle<()>>,
     sender: Sender<Event>,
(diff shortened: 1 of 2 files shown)


src/net/connection/timeout_thread.rs
```diff
@@ -2,6 +2,7 @@ use super::ConnectionPool;
 use error::{NetworkError, NetworkErrorKind, NetworkResult};
 use events::Event;
 
+use std::sync::atomic::*;
```
TimonPost: Could you please specify the types you are using here instead of `*` since we do that in the whole crate.

Philpax: Ah, good catch! I did that as part of my initial workaround, let me clean that up :)

src/net/connection/timeout_thread.rs
```diff
@@ -19,6 +20,7 @@ pub const TIMEOUT_POLL_INTERVAL: u64 = 1;
 /// 3. Check if the last time we have heard from them (received a packet from them) is greater than the amount of time considered to be a timeout.
 /// 4. If they have timed out, send a notification up the stack.
 pub struct TimeoutThread {
+    alive: Arc<AtomicBool>,
```
LucioFranco: Can you rename this to `shutdown_signal`?

src/net/connection/timeout_thread.rs
```diff
@@ -83,3 +96,10 @@ impl TimeoutThread {
         Ok(())
     }
 }
+
+impl Drop for TimeoutThread {
+    fn drop(&mut self) {
+        // Stop thread, and discard the error if it exists as it is irrelevant.
+        self.stop().ok();
```
LucioFranco: Can you log a `error!` here if the `stop` fails. That way the end user is notified in a best effort way.

src/net/connection/timeout_thread.rs
Philpax: Done.

src/net/connection/timeout_thread.rs
Philpax: Done.

PR #125, opened 2018-12-03
Fix up the benchmarks so that they run

jstnlef:
I'm not sure that this is even wanted or needed but it was just something I noticed when digging though the code.

### DIFF
--- a/benches/packet_processing.rs
+++ b/benches/packet_processing.rs
@@ -1,15 +1,13 @@
+extern crate byteorder;
 extern crate laminar;
 #[macro_use]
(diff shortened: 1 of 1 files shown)


TimonPost:
Didn't the benchmarks ran?

jstnlef:
The benchmarks weren't running for me. It seems that some of the structs the benchmarks were using were no longer publicly available.

TimonPost:
bord r=TimonPost

TimonPost:
bors r=TimonPost

Issue #108, opened 2018-11-13
Remove extern crates


TimonPost:
Currently we have dependencies to serde and bincode those are only used in examples. We need to fix those by for example putting the exanples in their own project. Please suggest some alternatives.

TimonPost:
Solved #126

PR #126, opened 2018-12-03
remove bincode and serde from lib.rs

jstnlef:
These libraries aren't being used anywhere but the examples so no need to declare them here.
Should address https://github.com/amethyst/laminar/issues/108

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -45,7 +45,6 @@
 
 #![warn(missing_docs)]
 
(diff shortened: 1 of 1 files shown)


TimonPost:
This probably broke the examples. They are using serde and bincode.

jstnlef:
The examples have their own `extern crate <package>` lines so they should still work fine 👍

PR #127, opened 2018-12-03
Replace methods which take &Arc with an Arc

jstnlef:
This gives some control back to the callers of these methods in terms of whether or not the Arc should be cloned.

### DIFF
--- a/src/events.rs
+++ b/src/events.rs
@@ -40,7 +40,7 @@ mod test {
 
         let test_conn = Arc::new(RwLock::new(VirtualConnection::new(
             addr.next().unwrap(),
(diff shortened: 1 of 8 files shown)


TimonPost:
Not sure if this is what I wanted. I used references to Arc with purpose. When using a reference to Ark the we don't  have to clone it our selfs everytime a function is called with Arc as parameter. The thing I saw happening a lot is that a lot of unnecessary  clones are done when a function takes in a Arc. By using a reference you don't need to clone until it is really needed.  Also for calling code it can  be cleaner.

TimonPost:
bors r=TimonPost

PR #129, opened 2018-12-05
Fix the benchmarks part 2

jstnlef:
Whoops. Since I changed the `new` method of `VirtualConnection` in another PR and these were merged out of order, more changes are needed.


TimonPost:
bors r=TimonPost

PR #132, opened 2018-12-08
Make 'recv()' function take an immutable reference to`self`

jstnlef:
Allow the caller to receive data on the socket without owning a mut reference. Addresses https://github.com/amethyst/laminar/issues/115.
Since the socket's `recv_buffer` is owned and only written to by the `recv()` method, we can use `Mutex` to provide that mutability.

### DIFF
--- a/examples/server_client.rs
+++ b/examples/server_client.rs
@@ -11,7 +11,7 @@ use laminar::{config::NetworkConfig, error::Result, net::UdpSocket, Packet};
 const SERVER: &str = "localhost:12351";
 
 fn server() -> Result<()> {
-    let mut socket = UdpSocket::bind(SERVER, NetworkConfig::default())?;
+    let socket = UdpSocket::bind(SERVER, NetworkConfig::default())?;
 
     println!("Listening for connections to {}", SERVER);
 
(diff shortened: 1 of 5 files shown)


jstnlef:
Locking and unlocking a Mutex is much quicker than allocating a new Vec each received packet. I'm not certain there is another solution. You definitely don't want the Mutex on the socket itself and the vec needs to be locked until the buffer is processed.

TimonPost:
Agree that it is probably faster. However maybe we could accomplish it like the original api. There is out have to pass a mutable buffer. That way you don't have to make it mutable. Only some redisgn is needed. 

I would suggest we run the benchmarks we have with mutex and without mutex. And see what the results are.

jstnlef:
I ran the benchmarks on my machine last night because I was curious (with master, then with the change). There was no notable difference on my machine at least. That said, I'm not sure the benchmarks are currently sufficient to really be able to pick up a difference.

LucioFranco:
Rebased against master I've fixed CI, thanks.

jstnlef:
Closing this since it'll likely become less useful in the future

Issue #134, opened 2018-12-29
high throughput reliability


Byter09:
Hello everyone.

I'm writing a small tool to verify this paper: https://arxiv.org/pdf/1302.1479.pdf
The implementation is pretty easy. There is a master distributing jobs and several slaves receiving jobs and sending back results.
I used/abused `laminar` for my usecase, because UDP seemed like the perfect choice for sending whole messages and all the fancy features like message-splitting and reliability were just the icing on top. It also made using `bincode` very easy and reliable.

I put my tool online so people can look at the code: https://gitlab.com/byter/csr-distributed
You can start a master by running `cargo run -- 3` and connect a slave by running `cargo run -- localhost:42111`. Be warned, there are a lot of debugging messages in there so it might be a bit spam-y. You can change the logging level by appending a `info` parameter or whatever level you want. Also no threading (a slave for each core) or clean-up happened yet - I hope this is excusable :D

Now to the actual issue:
When running a `master` with higher integers (for example the paper is for `9`, which would take weeks with a single slave) the job amount rises exponentially. This of course causes transmission rates to be way higher.
Using `5` alone causes it to return
https://github.com/amethyst/laminar/blob/b9564ffeb5b56e020767246a5536a6cadb407504/src/net/connection/virtual_connection.rs#L94

@TimonPost suggested on discord, this might be due to the sheer amount of data being transmitted and packets being dropped - including the header. So I investiged this a bit.

The last reported numbers by the slave were 
```
SENT PCKTS: sum: 33468 | pps: 293 | record pps: 2802
RECV PCKTS: sum: 33467 | pps: 292 | record pps: 2801

SENT BYTES: sum: 3301537 | bps: 29211 | record bps: 278766
RECV BYTES: sum: 1335351 | bps: 11795 | record bps: 112245
```
before it returned that error.

The corresponding master numbers:
```
SENT PCKTS: sum: 33467 | pps: 292 | record pps: 2801
RECV PCKTS: sum: 33467 | pps: 292 | record pps: 2801

SENT BYTES: sum: 1803889 | bps: 15883 | record bps: 151459
RECV BYTES: sum: 2832901 | bps: 25025 | record bps: 239454
```

Note: `bps` stands for `bytes per second` and  `pps` for `packets per second`
These values also directly come from `socket.recv()` and `socket.send()` and sending is done using the `ReliableUnordered` delivery method (because ReliableOrdered is not yet available).

[...]

PR #138, opened 2019-01-30
Use Mio socket rather than a blocking socket from the std library

jstnlef:
Based on what @TimonPost mentioned the other day, I took another look at how I could put together a minimum PR for the changes I made to my own re-implementation of Laminar using a mio socket while cleaning up the mutexes. This is that minimum PR. That said, a few of the naming changes did make it in because I believe that the present names are misleading. However, some of the larger restructuring can be held off for another time.
One of the interesting upshots I've noticed is that I've yet to run into [this issue](https://github.com/amethyst/laminar/issues/134) running the same test.


src/net/socket.rs
```diff
@@ -0,0 +1,222 @@
...
+        let poll = Poll::new()?;
+
+        poll.register(self, SOCKET, Ready::readable(), PollOpt::edge())?;
+
+        let mut events = Events::with_capacity(self.config.socket_event_buffer_size);
+        let events_ref = &mut events;
+        // Packet receiver MUST only be used in this method.
+        let packet_receiver = mem::replace(&mut self.packet_receiver, mpsc::channel().1);
+        // Nothing should break out of this loop!
+        loop {
```
TimonPost: Is there a way to stop the polling or are we even supposed to stop the polling?

src/net/socket.rs
```diff
@@ -0,0 +1,222 @@
...
+
+        let received_payload = &self.recv_buffer[..recv_len];
+        let connection = self
+            .connections
+            .get_or_insert_connection(&address, self.config.clone());
+        connection.process_incoming(received_payload)
+    }
+
+    /// Send a single packet over the udp socket.
+    fn send_packet(&self, addr: &SocketAddr, payload: &[u8]) -> NetworkResult<usize> {
```
TimonPost: I prefer to have the same method name as the function what it is wrapping: `send_to`

src/net/socket.rs
```diff
@@ -0,0 +1,222 @@
...
+                event_sender,
+                packet_receiver,
+            },
+            packet_sender,
+            event_receiver,
+        )
+    }
+}
+
+impl Evented for LaminarSocket {
```
TimonPost: Shouldn't this be documented: `register`,  `reregister` and `deregister`  because they are callable from the Api

src/net/socket.rs
```diff
@@ -0,0 +1,222 @@
...
+
+        for address in idle_addresses {
+            self.connections.remove_connection(&address);
+            if let Err(err) = self.event_sender.send(SocketEvent::TimeOut(address)) {
+                error!("{:?}", err);
+            }
+        }
+    }
+
+    /// Process events received from the mio socket.
```
LucioFranco: can just be `//` since this is not public

src/net/socket.rs
```diff
@@ -0,0 +1,222 @@
...
+                        }
+                    }
+                }
+                _ => unreachable!(),
+            }
+        }
+        Ok(())
+    }
+
+    /// Serializes and sends a `Packet` on the socket. On success, returns the number of bytes written.
```
LucioFranco: same here

src/net/socket.rs
LucioFranco: they are documented here https://docs.rs/mio/0.6.16/mio/event/trait.Evented.html so no need

src/net/socket.rs
```diff
@@ -0,0 +1,222 @@
...
+use std::mem;
+use std::net::{SocketAddr, ToSocketAddrs};
+use std::sync::{mpsc, Arc};
+
+const SOCKET: Token = Token(0);
+
+/// A reliable UDP socket implementation with configurable reliability and ordering guarantees.
+pub struct LaminarSocket {
+    socket: mio::net::UdpSocket,
+    config: Arc<NetworkConfig>,
```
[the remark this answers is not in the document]
TimonPost: Yes, as long it is used at more places it is better to do so. We did this because we were making multiple copies of it.

src/net/socket.rs
```diff
@@ -0,0 +1,222 @@
...
+use mio::{Evented, Events, Poll, PollOpt, Ready, Token};
+use std::io;
+use std::mem;
+use std::net::{SocketAddr, ToSocketAddrs};
+use std::sync::{mpsc, Arc};
+
+const SOCKET: Token = Token(0);
+
+/// A reliable UDP socket implementation with configurable reliability and ordering guarantees.
+pub struct LaminarSocket {
```
LucioFranco: As discussed on discord this should be renamed to `Socket` with also `NetworkConfig` -> `Config`. `Socket` can still live under net but can be reexported at the root of the project so that people can access them as `laminar::{Socket, Config}`.

src/net/socket.rs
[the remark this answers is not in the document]
LucioFranco: right what im saying is we should analyze to see if we even need this as it may just be uneeded overhead.

src/config.rs
```diff
@@ -35,18 +37,25 @@ pub struct NetworkConfig {
...
+            idle_connection_timeout: Duration::from_secs(5),
             max_packet_size: (MAX_FRAGMENTS_DEFAULT * FRAGMENT_SIZE_DEFAULT) as usize,
             max_fragments: MAX_FRAGMENTS_DEFAULT as u8,
             fragment_size: FRAGMENT_SIZE_DEFAULT,
             fragment_reassembly_buffer_size: 64,
             receive_buffer_max_size: 1500,
             rtt_smoothing_factor: 0.10,
             rtt_max_value: 250,
+            socket_event_buffer_size: 1024,
+            socket_polling_timeout: Some(Duration::from_millis(100)),
```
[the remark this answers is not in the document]
jstnlef: This timeout is the how long the poll method blocks while waiting for a packet to process.

src/net/socket.rs
jstnlef: That's a good question. I mean, if we stop polling we'll stop receiving packets but perhaps it's useful to have.

src/config.rs
```diff
@@ -1,9 +1,11 @@
 use crate::net::constants::{FRAGMENT_SIZE_DEFAULT, MAX_FRAGMENTS_DEFAULT};
-use std::default::Default;
+use std::{default::Default, time::Duration};
 
 #[derive(Clone)]
 /// Struct that contains config values for various aspects of the network
 pub struct NetworkConfig {
+    /// The maximal amount of time to keep `VirtualConnection`s around before cleaning them up.
+    pub idle_connection_timeout: Duration,
```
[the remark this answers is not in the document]
jstnlef: This is our own internal timeout for removing VirtualConnections. I'm inclined to leave it. Gives us some optionality.

src/net/socket.rs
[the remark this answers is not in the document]
jstnlef: I mostly wanted to minimize the changes so I left this. I did remove it in my refactor, however. I can remove it if we want.

src/net/connection.rs
```diff
@@ -1,17 +1,135 @@
...
+    }
+
+    /// Try to get a VirtualConnection by address. If the connection does not exist, it will be
+    /// inserted and returned.
+    pub fn get_or_insert_connection(
+        &mut self,
+        address: &SocketAddr,
+        config: Arc<NetworkConfig>,
+    ) -> &mut VirtualConnection {
+        if !self.connections.contains_key(address) {
```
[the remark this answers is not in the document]
jstnlef: Ah that was the incantation I was looking for 👍

src/net/socket.rs
```diff
@@ -0,0 +1,222 @@
...
+    socket: mio::net::UdpSocket,
+    config: Arc<NetworkConfig>,
+    connections: ActiveConnections,
+    recv_buffer: Vec<u8>,
```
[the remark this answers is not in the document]
jstnlef: This was originally a `Vec` and I'm not entirely sure how I can have an array where the size is configurable at runtime.

src/net/socket.rs
```diff
@@ -0,0 +1,222 @@
...
+    event_sender: mpsc::Sender<SocketEvent>,
+    packet_receiver: mpsc::Receiver<Packet>,
+}
+
+impl LaminarSocket {
+    /// Binds to the socket and then sets up `ActiveConnections` to manage the "connections".
+    /// Because UDP connections are not persistent, we can only infer the status of the remote
+    /// endpoint by looking to see if they are still sending packets or not
+    pub fn bind<A: ToSocketAddrs>(
+        addresses: A,
```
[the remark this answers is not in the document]
jstnlef: Hrm. I'm curious as to how it's a blocking call. Or, more to the point, can it be particularly slow?

src/net/socket.rs
```diff
@@ -0,0 +1,222 @@
...
+        config: NetworkConfig,
+    ) -> NetworkResult<(Self, mpsc::Sender<Packet>, mpsc::Receiver<SocketEvent>)> {
+        let socket = std::net::UdpSocket::bind(addresses)?;
```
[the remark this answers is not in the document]
jstnlef: Well the reason is because mio::net::UdpSocket doesn't take a `ToSocketAddrs`

src/net/socket.rs
```diff
@@ -0,0 +1,222 @@
...
+}
+
+impl LaminarSocket {
+    /// Binds to the socket and then sets up `ActiveConnections` to manage the "connections".
+    /// Because UDP connections are not persistent, we can only infer the status of the remote
+    /// endpoint by looking to see if they are still sending packets or not
+    pub fn bind<A: ToSocketAddrs>(
+        addresses: A,
+        config: NetworkConfig,
+    ) -> NetworkResult<(Self, mpsc::Sender<Packet>, mpsc::Receiver<SocketEvent>)> {
```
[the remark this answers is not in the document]
jstnlef: That's not a bad idea. I can take a stab at it.

src/net/socket.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    /// Entry point to the run loop. This should run in a spawned thread since calls to `poll.poll`
+    /// are blocking.
+    pub fn start_polling(&mut self) -> NetworkResult<()> {
+        let poll = Poll::new()?;
+
+        poll.register(self, SOCKET, Ready::readable(), PollOpt::edge())?;
+
+        let mut events = Events::with_capacity(self.config.socket_event_buffer_size);
+        let events_ref = &mut events;
```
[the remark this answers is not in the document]
jstnlef: Sure no problem 👍

src/net/socket.rs
```diff
@@ -0,0 +1,222 @@
...
+        // Nothing should break out of this loop!
+        loop {
+            self.handle_idle_clients();
+            if let Err(e) = poll.poll(events_ref, self.config.socket_polling_timeout) {
+                error!("Error polling the socket: {:?}", e);
+            }
+            if let Err(e) = self.process_events(events_ref) {
+                error!("Error processing events: {:?}", e);
+            }
+            // XXX: I'm fairly certain this isn't exactly safe. I'll likely need to add some
```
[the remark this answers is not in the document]
jstnlef: There likely needs to be a check here to `io::ErrorKind::WouldBlock` if we try to send packets which are too large. However, we do a good amount of work to ensure that the packet sizes are small via fragmentation. Basically, will need to do some testing if this ends up being a problem

src/net/connection.rs
```diff
@@ -1,17 +1,135 @@
...
+use crate::config::NetworkConfig;
+use std::{collections::HashMap, net::SocketAddr, sync::Arc, time::Duration};
 
-pub type Connection = Arc<RwLock<VirtualConnection>>;
-pub type Connections = HashMap<SocketAddr, Connection>;
-pub type ConnectionsCollection = Arc<RwLock<Connections>>;
+/// Maintains a registry of active "connections". Essentially, when we receive a packet on the
+/// socket from a particular `SocketAddr`, we will track information about it here.
+pub struct ActiveConnections {
+    connections: HashMap<SocketAddr, VirtualConnection>,
```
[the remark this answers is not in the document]
LucioFranco: May want to think about using FnvHashmap or hashbrown though iirc Fnv has the best perf for a smaller hashmap.

src/net/connection.rs
```diff
@@ -1,17 +1,135 @@
...
+        address: &SocketAddr,
+        config: Arc<NetworkConfig>,
+    ) -> &mut VirtualConnection {
+        if !self.connections.contains_key(address) {
+            self.connections
+                .insert(*address, VirtualConnection::new(*address, config));
+        }
+        self.connections
+            .get_mut(address)
+            .expect("We just added this key. It should definitely exist.")
```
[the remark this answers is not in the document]
LucioFranco: Actually, on this note, I would rather have something along the lines of "This is a bug!" as that is clear to the user this is a bug.

src/net/socket.rs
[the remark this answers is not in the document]
LucioFranco: 👍 I suspect we can remove almost all these arcs since we sit within one thread now instead of multiple. So I think as a part of this change we should try and remove our dependence on `std::sync` as much as possible.

src/net/socket.rs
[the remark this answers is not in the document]
LucioFranco: Why should the size be configurable at runtime? What is the use case around that?

src/net/socket.rs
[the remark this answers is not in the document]
LucioFranco: Since, `ToSocketAddr::to_socket_addrs` which extracts the socket can block, most APIs at this level of the transport are not concerned with dns res so we should be explicit this is not something we provide. It blocks because it makes a syscall to just get the socketaddr. This is why mio doesn't accept it.

src/net/socket.rs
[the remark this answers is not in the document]
LucioFranco: Right, as my point above, I don't think we should accept a `ToSocketAddrs` 😄

src/net/socket.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// are blocking.
+    pub fn start_polling(&mut self) -> NetworkResult<()> {
+        let poll = Poll::new()?;
+
+        poll.register(self, SOCKET, Ready::readable(), PollOpt::edge())?;
+
+        let mut events = Events::with_capacity(self.config.socket_event_buffer_size);
+        let events_ref = &mut events;
+        // Packet receiver MUST only be used in this method.
+        let packet_receiver = mem::replace(&mut self.packet_receiver, mpsc::channel().1);
```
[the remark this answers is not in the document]
LucioFranco: `try_iter` requires `&self` so there shouldn't be any borrow issues, but we can try to isolate this into a play.rust-lang.org and see what we can work out. Generally speaking, using `mem` stuff should NEVER be used unless you are doing something really funky which this isn't. I would rather play the borrow checker game.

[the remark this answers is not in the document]
LucioFranco: feel free to ping me on discord about this and we can talk further.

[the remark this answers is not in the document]
jstnlef: I'm completely with you. I don't love the idea of using `mem`. However, the issue is that `self.packet_receiver.try_iter()` borrows `&self` and `self.send_to` requires a `&mut self`. We can talk about it more after I get home from work.

src/net/connection.rs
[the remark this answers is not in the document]
jstnlef: I've already replaced this with the code you suggested. I wish you could hide outdated comments here :P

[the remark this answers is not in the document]
LucioFranco: hit that resolve button 😉 
image

src/net/socket.rs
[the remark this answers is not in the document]
jstnlef: I'll do that tonight. 👍

src/net/socket.rs
jstnlef: I'm inclined to leave this.

src/net/socket.rs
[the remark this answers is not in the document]
TimonPost: It should not be configurable. This is just the buffer the packets are written to. When we do a receive with UDP we get the size back, which we can use for indexing into this buffer. A vector is fine here since it is created once with the size from the config.

[Here](https://github.com/amethyst/laminar/blob/a3cd32bb7cddffe0e43633798e1dbd903243ec35/src/net/socket.rs#L187) the buffer is created once.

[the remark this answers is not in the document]
jstnlef: I mean, I'm ok with locking it to the size of MTU (1500). However, if we do want it to be configurable, which it was originally, this must be a Vec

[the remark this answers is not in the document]
LucioFranco: If anything we don't need runtime configurable we need compile-time configurable buffer size.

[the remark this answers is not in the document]
fhaynes: Don't hard code the MTU please. It'll kill performance when using this for simulations in clusters where jumbo frames work fine, for example.

src/net/socket.rs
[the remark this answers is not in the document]
fhaynes:
```
            let mut packet_iter = self.packet_receiver.try_iter();
            for packet in packet_iter.next() {
                if let Err(e) = self.send_to(packet) {
                    error!("Error sending packet: {:?}", e);
                }
            }
```

src/net/socket.rs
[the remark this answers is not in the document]
jstnlef: Definitely think leaving this as a `Vec` is the right move for now.

src/net/socket.rs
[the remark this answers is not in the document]
fhaynes: So close! Sadly this didn't work. =(

benches/packet_processing.rs
```diff
@@ -22,11 +22,9 @@ fn process_packet_before_send(
...
 
 fn send_unreliable_benchmark(c: &mut Criterion) {
-    let config = NetworkConfig::default();
-    let mut connection = VirtualConnection::new(
-        SERVER_ADDR.parse().unwrap(),
-        Arc::new(NetworkConfig::default()),
-    );
+    let config = Config::default();
+    let mut connection =
+        VirtualConnection::new(SERVER_ADDR.parse().unwrap(), Arc::new(Config::default()));
```
fhaynes: Unwrap alert! Unwrap alert!

Can we use expect?

benches/packet_processing.rs
```diff
@@ -40,11 +38,9 @@ fn send_unreliable_benchmark(c: &mut Criterion) {
...
 
 fn send_reliable_benchmark(c: &mut Criterion) {
-    let config = NetworkConfig::default();
-    let mut connection = VirtualConnection::new(
-        SERVER_ADDR.parse().unwrap(),
-        Arc::new(NetworkConfig::default()),
-    );
+    let config = Config::default();
+    let mut connection =
+        VirtualConnection::new(SERVER_ADDR.parse().unwrap(), Arc::new(Config::default()));
```
fhaynes: Same unwrap alert

benches/packet_processing.rs
```diff
@@ -84,10 +80,8 @@ fn acked_header_bytes(
 }
 
 fn receive_unreliable_benchmark(c: &mut Criterion) {
-    let mut connection = VirtualConnection::new(
-        SERVER_ADDR.parse().unwrap(),
-        Arc::new(NetworkConfig::default()),
-    );
+    let mut connection =
+        VirtualConnection::new(SERVER_ADDR.parse().unwrap(), Arc::new(Config::default()));
```
fhaynes: Same unwrap alert

benches/packet_processing.rs
```diff
@@ -99,10 +93,8 @@ fn receive_unreliable_benchmark(c: &mut Criterion) {
 }
 
 fn receive_reliable_benchmark(c: &mut Criterion) {
-    let mut connection = VirtualConnection::new(
-        SERVER_ADDR.parse().unwrap(),
-        Arc::new(NetworkConfig::default()),
-    );
+    let mut connection =
+        VirtualConnection::new(SERVER_ADDR.parse().unwrap(), Arc::new(Config::default()));
```
fhaynes: Same unwrap alert.

examples/server_client.rs
```diff
@@ -3,18 +3,20 @@
 //! Note that in practice you don't want to implement a chat client using UDP.
 use std::io::stdin;
 
-use laminar::{config::NetworkConfig, error::Result, net::UdpSocket, Packet};
+use laminar::{Config, NetworkError, Packet, Socket, SocketEvent};
+use std::thread;
 
-const SERVER: &str = "localhost:12351";
+const SERVER: &str = "127.0.0.1:12351";
```
fhaynes: Just out of curiosity, why the change from localhost?

examples/server_client.rs
```diff
@@ -3,18 +3,20 @@
...
+fn server() -> Result<(), NetworkError> {
+    let (mut socket, packet_sender, event_receiver) = Socket::bind(SERVER, Config::default())?;
+    let _thread = thread::spawn(move || socket.start_polling());
 
     println!("Listening for connections to {}", SERVER);
 
     loop {
-        match socket.recv()? {
-            Some(packet) => {
+        match event_receiver.recv().expect("Should get a message") {
```
fhaynes: This will kill the application if any error is encountered. Is that what we want?

fhaynes: And yes, I know this is in the examples, but I think our examples should probably be as close to production as possible.

src/config.rs
```diff
@@ -1,9 +1,11 @@
 use crate::net::constants::{FRAGMENT_SIZE_DEFAULT, MAX_FRAGMENTS_DEFAULT};
-use std::default::Default;
+use std::{default::Default, time::Duration};
 
 #[derive(Clone)]
 /// Struct that contains config values for various aspects of the network
-pub struct NetworkConfig {
+pub struct Config {
+    /// The maximal amount of time to keep `VirtualConnection`s around before cleaning them up.
```
fhaynes: Probably wanted an apostrophe here?

src/net/connection.rs
```diff
@@ -1,17 +1,129 @@
...
+
+    /// Removes the connection from `ActiveConnections` by socket address.
+    pub fn remove_connection(
+        &mut self,
+        address: &SocketAddr,
+    ) -> Option<(SocketAddr, VirtualConnection)> {
+        self.connections.remove_entry(address)
+    }
+
+    /// Check for and return `VirtualConnection`s which have been idling longer than `max_idle_time`.
```
fhaynes: Meant an apostrophe again?

src/net/connection.rs
```diff
@@ -1,17 +1,129 @@
...
+impl ActiveConnections {
+    pub fn new() -> Self {
+        Self {
+            connections: HashMap::new(),
+        }
+    }
+
+    /// Try to get a VirtualConnection by address. If the connection does not exist, it will be
+    /// inserted and returned.
+    pub fn get_or_insert_connection(
```
fhaynes: This is now blocking, correct? In that while we are adding a new connection, we can't process incoming data?

src/net/connection.rs
```diff
@@ -1,17 +1,129 @@
...
+    pub fn idle_connections(&mut self, max_idle_time: Duration) -> Vec<SocketAddr> {
+        self.connections
+            .iter()
+            .filter(|(_, connection)| connection.last_heard() >= max_idle_time)
+            .map(|(address, _)| *address)
+            .collect()
+    }
+
+    /// Get the number of connected clients.
+    #[allow(dead_code)]
```
fhaynes: Do we still need this allow?

examples/server_client.rs
LucioFranco: localhost I think requires to resolve and address can be converted to a `SocketAddr`.

benches/packet_processing.rs
jstnlef: is this code even run? Yeah, look like it's not run at all because it would be completely broken.

src/net/connection.rs
LucioFranco: its not blocking because there is no lock being acquired since no synchronization is required 😄

src/net/connection.rs
LucioFranco: I think this was added early on for something but we haven't used it so I think it is fair to remove this function or at least this warning.

examples/server_client.rs
jstnlef: Was getting an error if I tried to bind to localhost.

examples/server_client.rs
jstnlef: That's a good question. I defer to the group on this one.

src/config.rs
jstnlef: Wrapping in back ticks produces different markdown output so it's signifying a specific type. `VirtualConnection`

src/net/connection.rs
jstnlef: With the async socket, there's only 1 thread reading, managing connections, and writing so technically speaking, when we are in this piece of code we are not actively reading from the socket but the OS is buffering up received data when we get to the reading phase of the loop.

src/net/connection.rs
jstnlef: Method is used in test code.

src/net/socket.rs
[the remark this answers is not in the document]
jstnlef: Added a `from_std` method to bind to the socket.

TimonPost:
It does not seem to be working on windows. 

I ran this code: https://gist.github.com/TimonPost/017adff60c860d7ec617e21cf7c09606

_Cargo.toml_

```
[[example]]
name = "sc"
path = "examples/server_client.rs"
```

_start the server_
`cargo run --example sc` and type 'server' when the program asks you for input

_start the client_ 
`cargo run --example sc` and type 'client' when the program asks you for input.

jstnlef:
@TimonPost Looks like you forgot `let _thread = thread::spawn(move || socket.start_polling());` line in the client.

src/net/socket.rs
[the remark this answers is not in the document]
LucioFranco: I've opened an issue regarding this #141

src/net/socket.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+
+impl Socket {
+    /// Binds to the socket and then sets up `ActiveConnections` to manage the "connections".
+    /// Because UDP connections are not persistent, we can only infer the status of the remote
+    /// endpoint by looking to see if they are still sending packets or not
+    pub fn bind<A: ToSocketAddrs>(
+        addresses: A,
+        config: Config,
+    ) -> NetworkResult<(Self, Sender<Packet>, Receiver<SocketEvent>)> {
```
LucioFranco: For future reference we should probably move these to some sort of `Socket` and `SocketHandle` where `SocketHandle` is clonable and provides the send and recv interface.  This would then allow you to use `Socket` to drive the inner transport and the `SocketHandle` to read and write. That should clean this up. `Socket` should still provide the read and write interface but is not clonable.

This is what Im thinking

```rust
struct Socket {
	fn send(&mut self, packet: Packet) -> Result<()>;
	fn recv(&mut self) -> Packet;
	fn start_polling(&mut self) -> Result<()>;
}

#[derive(Clone)]
struct SocketHandle {
	fn send(&mut self, packet: Packet) -> Result<()>;
	fn recv(&mut self) -> Packet;
}
```

mostly this is just me thinking out loud would love to hear more thoughts on this

src/packet/header/fragment_header.rs
```diff
@@ -53,10 +53,10 @@ impl FragmentHeader {
     }
 }
 
-impl HeaderParser for FragmentHeader {
+impl HeaderWriter for FragmentHeader {
     type Output = NetworkResult<()>;
 
-    fn parse(&self, buffer: &mut Vec<u8>) -> <Self as HeaderParser>::Output {
+    fn parse(&self, buffer: &mut Vec<u8>) -> <Self as HeaderWriter>::Output {
```
LucioFranco: This return type should be `Self::Output` iirc

src/net/socket.rs
```diff
@@ -0,0 +1,229 @@
...
+            .connections
+            .get_or_insert_connection(address, &self.config);
+        connection.process_incoming(received_payload)
+    }
+
+    // Send a single packet over the UDP socket.
+    fn send_packet(&self, addr: &SocketAddr, payload: &[u8]) -> NetworkResult<usize> {
+        let bytes_sent = self
+            .socket
+            .send_to(payload, addr)
```
LucioFranco: We don't actually poll to see if this is ready to write too so we should understand what the behvaior is here.

From our discord discussion with @jstnlef:

```
jstnlefToday at 11:18 PM
that won't block
but it "may" error with a "WouldBlock"
if the packet you're trying to send is too large
```

From my understanding, this means this will always write and return right away unless the packet is too large. We should understand the error cases here or buffer and request to write.

src/net/socket.rs
```diff
@@ -0,0 +1,229 @@
...
+                                    {
+                                        break;
+                                    }
+                                    _ => error!("Error receiving from socket: {:?}", err),
+                                },
+                            };
+                        }
+                    }
+                }
+                _ => unreachable!(),
```
LucioFranco: I think we can add a string here to be more clear if this ever does hit for some odd reason. Better than just an unreachable panic.

src/net/socket.rs
LucioFranco: to add to this it seems we do not handle the `WouldBlock` case, this should always be handled for proper portability https://docs.rs/mio/0.6.16/mio/struct.Poll.html#spurious-events

src/net/socket.rs
jstnlef: Actually I was thinking about this. Can do this now or in a future PR.

src/packet/header/fragment_header.rs
jstnlef: Again, one of those attempts to minimize the code change. I had fixed this in my rebuild.

LucioFranco: 👍 if we can fix this in this PR that would be nice if not I dont mind

src/net/socket.rs
LucioFranco: let's do future, this was mostly me just dumping the thoughts out of my brain.

LucioFranco: I opened an issue for this #142

src/packet/header/fragment_header.rs
jstnlef: 👍

src/net/local_ack.rs
```diff
@@ -13,11 +13,13 @@ pub struct LocalAckRecord {
 
 impl LocalAckRecord {
     /// Checks if there are packets in the queue to be acknowledged.
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
jstnlef: This is used in all of the unit tests.

src/net/socket.rs
jstnlef: Remind me where we ended up on this one?

LucioFranco: After discussion I think if send_to fails we need to notify the user that the send did not succeed.

src/net/local_ack.rs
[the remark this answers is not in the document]
LucioFranco: then we probably want to add some sort of cfg attribute what do you think?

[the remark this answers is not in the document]
jstnlef: cfg attribute? For what reason?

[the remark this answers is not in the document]
LucioFranco: something like `#[cfg(test)]`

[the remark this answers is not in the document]
jstnlef: oh right! I completely forgot about that 👍

TimonPost:
RE: Windows Issue.

knipsel

As you can see, the left one - the server - only receives 1 packet/event a second and the right one - the client sents 100 a second.

Replace `socket.rs` for a nice debugging output to [this](https://gist.github.com/TimonPost/9125b6680dbb98c691ce14360813d173).
Replace `server_client.rs` with [this](https://gist.github.com/TimonPost/daf2d1cdf3d03628edde4057beda419a)

_Cargo.toml_

```
[[example]]
name = "sc"
path = "examples/server_client.rs"
```

_start the server_
`cargo run --example sc` and type 'server' when the program asks you for input

_start the client_ 
`cargo run --example sc` and type 'client' when the program asks you for input.

fhaynes:
bors r=timonpost,jstnlef

fhaynes:
bors r=timonpost,fhaynes

fhaynes:
bors r=timonpost,fhaynes

PR #149, opened 2019-03-08
Multiplexing

fhaynes:
This changes the socket handling to not use mio until the Windows issue is resolved.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 # Set up the Rust toolchain.
 language: rust
+
+env:
+  - RUST_LOG="debug"
+
 rust:
   - stable
   - beta
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,6 @@
 # Set up the Rust toolchain.
 language: rust
 
-env:
-  - RUST_LOG="debug"
-
 rust:
   - stable
   - beta
--- a/benches/packet_processing.rs
+++ b/benches/packet_processing.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
 
-use laminar::{net::VirtualConnection, Config, DeliveryMethod, ProtocolVersion};
+use laminar::{Config, DeliveryMethod, ProtocolVersion, VirtualConnection};
 
 use criterion::{criterion_group, criterion_main, Criterion};
 
@@ -19,34 +19,6 @@ fn process_packet_before_send(
     let _packet_data = connection
         .process_outgoing(&payload, delivery_method)
         .unwrap();
-}
-
-fn send_unreliable_benchmark(c: &mut Criterion) {
-    let config = Config::default();
-    let mut connection =
-        VirtualConnection::new(SERVER_ADDR.parse().unwrap(), &Arc::new(Config::default()));
-
-    c.bench_function("process unreliable before send", move |b| {
-        b.iter(|| {
-            process_packet_before_send(
-                &mut connection,
-                &config,
-                DeliveryMethod::UnreliableUnordered,
-            )
-        })
-    });
-}
-
-fn send_reliable_benchmark(c: &mut Criterion) {
-    let config = Config::default();
-    let mut connection =
-        VirtualConnection::new(SERVER_ADDR.parse().unwrap(), &Arc::new(Config::default()));
-
-    c.bench_function("process reliable before send", move |b| {
-        b.iter(|| {
-            process_packet_before_send(&mut connection, &config, DeliveryMethod::ReliableUnordered)
-        })
-    });
 }
 
 fn process_packet_when_received(connection: &mut VirtualConnection, data: &[u8]) {
@@ -107,8 +79,6 @@ fn receive_reliable_benchmark(c: &mut Criterion) {
 
 criterion_group!(
     benches,
-    send_reliable_benchmark,
-    send_unreliable_benchmark,
     receive_unreliable_benchmark,
     receive_reliable_benchmark
 );
--- a/benches/packet_processing.rs
+++ b/benches/packet_processing.rs
@@ -24,7 +24,7 @@ fn process_packet_before_send(
 fn send_unreliable_benchmark(c: &mut Criterion) {
     let config = Config::default();
     let mut connection =
-        VirtualConnection::new(SERVER_ADDR.parse().unwrap(), Arc::new(Config::default()));
+        VirtualConnection::new(SERVER_ADDR.parse().unwrap(), &Arc::new(Config::default()));
 
     c.bench_function("process unreliable before send", move |b| {
         b.iter(|| {
@@ -40,7 +40,7 @@ fn send_unreliable_benchmark(c: &mut Criterion) {
 fn send_reliable_benchmark(c: &mut Criterion) {
     let config = Config::default();
     let mut connection =
-        VirtualConnection::new(SERVER_ADDR.parse().unwrap(), Arc::new(Config::default()));
+        VirtualConnection::new(SERVER_ADDR.parse().unwrap(), &Arc::new(Config::default()));
 
     c.bench_function("process reliable before send", move |b| {
         b.iter(|| {
@@ -81,7 +81,7 @@ fn acked_header_bytes(
 
 fn receive_unreliable_benchmark(c: &mut Criterion) {
     let mut connection =
-        VirtualConnection::new(SERVER_ADDR.parse().unwrap(), Arc::new(Config::default()));
+        VirtualConnection::new(SERVER_ADDR.parse().unwrap(), &Arc::new(Config::default()));
 
     // setup fake received bytes.
     let mut buffer = standard_header_bytes(DeliveryMethod::UnreliableUnordered);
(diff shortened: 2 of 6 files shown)


src/lib.rs
```diff
@@ -20,29 +20,6 @@
 //!
 //! # Example
 //!
-//! ```rust
```
LucioFranco: why remove this?

src/net/socket.rs
```diff
@@ -5,18 +5,17 @@ use crate::{
     packet::Packet,
 };
 use crossbeam_channel::{self, unbounded, Receiver, Sender};
-use log::error;
-use mio::{Evented, Events, Poll, PollOpt, Ready, Token};
+#[allow(unused_imports)]
```
LucioFranco: can we just remove the imports? 😄

src/net/socket.rs
```diff
@@ -33,100 +32,81 @@ impl Socket {
...
-                // so maybe it's work switching to that while providing the same interface?
-                if let Err(e) = self.send_to(packet) {
-                    error!("Error sending packet: {:?}", e);
+            // First we pull any newly arrived packets and handle them
+            match self.recv_from() {
+                Ok(result) => match result {
+                    Some(packet) => {
+                        match self.event_sender.send(SocketEvent::Packet(packet)) {
+                            Ok(p) => {
+                                println!("Packet received: {:?}", p);
```
LucioFranco: im assuming this is the wip part?

tests/common.rs
```diff
@@ -62,7 +58,8 @@ impl ServerMoq {
                     packet_throughput = 0;
                 }
             }
-        })
+        });
```
LucioFranco: take it or leave it: remove the semi and you dont need to `let handle`

src/lib.rs
fhaynes: Because it fails for some reason with the new architecture, though all other tests pass. We'll have to write a new example doctest.

LucioFranco: 👍 we should prob add a `todo` line

src/lib.rs
```diff
@@ -56,7 +33,7 @@ mod config;
 /// All internal error handling logic
 mod error;
 /// Networking modules
-mod net;
+pub mod net;
```
TimonPost: do we want this to be public? I thought we re-exported all types in `lib.rs`

tests/fragments.rs
```diff
@@ -1,13 +1,14 @@
 mod common;
 
+use crate::common::{ClientStub, ServerMoq};
 use laminar::{Config, DeliveryMethod};
+#[allow(unused_imports)]
+use log::{debug, error, info};
```
TimonPost: removing? Probably done because of this WIP right?.

src/net/socket.rs
fhaynes: I'm fine with that, they are just there for logging purposes. I go back and forth between only importing the ones used, and importing them all to save future people the trouble. I dunno. I'll prolly just remove them.

src/net/socket.rs
fhaynes: Why, uh, yes, yes it is.

src/net/socket.rs
LucioFranco: I actually have been prefering to just `macro_use` them since they are so common and are only a few macros that everyone knows.

src/lib.rs
fhaynes: For some reason, it threw an error. I'll see if I can figure out how to do it without making that public.

tests/fragments.rs
fhaynes: Yes

src/lib.rs
fhaynes: An issue would probably be more effective. No one reads TODO lines. =)

tests/common.rs
fhaynes: There was a reason for that, but I can't remember it now, and it works fine without it.

src/lib.rs
fhaynes: Huh, works without it now. Ok then.

src/net/socket.rs
fhaynes: Example please?

benches/packet_processing.rs
```diff
@@ -24,7 +24,7 @@ fn process_packet_before_send(
 fn send_unreliable_benchmark(c: &mut Criterion) {
     let config = Config::default();
     let mut connection =
-        VirtualConnection::new(SERVER_ADDR.parse().unwrap(), Arc::new(Config::default()));
+        VirtualConnection::new(SERVER_ADDR.parse().unwrap(), &Arc::new(Config::default()));
```
jstnlef: Looking at this now, it looks like we either need to rewrite these benches or delete them completely. They don't run in CI and are broken when run with `cargo bench`.

src/net/socket.rs
```diff
@@ -33,47 +30,63 @@ impl Socket {
...
+                        error!("Empty packet received");
+                    }
+                },
+                Err(e) => {
+                    error!("Error receiving packet: {:?}", e);
+                }
+            };
+
+            // Now grab all the packets waiting to be sent and send them
+            let to_send: Vec<Packet> = self.packet_receiver.try_iter().collect();
```
jstnlef: Allocation per loop iteration isn't great.

src/net/socket.rs
```diff
@@ -159,29 +132,37 @@ impl Socket {
...
+                    return Err(ErrorKind::ReceivedDataToShort)?;
+                }
+                let received_payload = &self.recv_buffer[..recv_len];
+                let connection = self
+                    .connections
+                    .get_or_insert_connection(address, &self.config);
+                connection.process_incoming(received_payload)
+            }
+            Err(ref e) if e.kind() == io::ErrorKind::WouldBlock => {
+                error!("Encountered a WouldBlock: {:?}", e);
```
jstnlef: This should likely just return if the socket would have been blocked. Also, we probably don't want to lose any of the packets pulled off of the channel in the caller so it may make sense to bubble up the error and handle it at the call site.

src/net/socket.rs
fhaynes: oh, thanks, meant to do that.

benches/packet_processing.rs
fhaynes: Deleting them. We can write new ones after @TimonPost 's PR and this are merged

src/net/socket.rs
fhaynes: Sorry, not sure what you mean. If it hits the WouldBlock arm, it will return. If we want it to bubble up the error, we'll have to wrap the socket::io error types with our own error type.

jstnlef: You know what, I take it back. I believe I was mistaking this for the send method. This is fine.

TimonPost:
I think that this works good, can we merge this?

jstnlef:
@TimonPost This is still marked as a WIP and as far as I'm aware, not actually done yet.

TimonPost:
I ran the code seems to work, I know it is a WIP, and that there is still some code that should be removed, @fhaynes??

src/net/socket.rs
```diff
@@ -33,47 +30,64 @@ impl Socket {
...
+                        error!("Empty packet received");
+                    }
+                },
+                Err(e) => {
+                    error!("Error receiving packet: {:?}", e);
+                }
+            };
+
+            // Now grab all the packets waiting to be sent and send them
+            to_send = self.packet_receiver.try_iter().collect();
```
jstnlef: This is still doing an allocation per loop iteration. Basically, any time you call `collect()` you're making a new `Vec`

src/net/socket.rs
```diff
@@ -159,29 +133,37 @@ impl Socket {
...
-
         Ok(bytes_sent)
     }
 
-    fn new(
-        socket: mio::net::UdpSocket,
-        config: Config,
-    ) -> (Self, Sender<Packet>, Receiver<SocketEvent>) {
+    #[allow(dead_code)]
+    fn new(socket: UdpSocket, config: Config) -> (Self, Sender<Packet>, Receiver<SocketEvent>) {
```
jstnlef: Should probably remove this if it's not being used.

src/net/socket.rs
```diff
@@ -159,70 +128,31 @@ impl Socket {
...
+                    return Err(ErrorKind::ReceivedDataToShort)?;
+                }
+                let received_payload = &self.recv_buffer[..recv_len];
+                let connection = self
+                    .connections
+                    .get_or_insert_connection(address, &self.config);
+                connection.process_incoming(received_payload)
+            }
+            Err(ref e) if e.kind() == io::ErrorKind::WouldBlock => {
+                error!("Encountered a WouldBlock: {:?}", e);
```
TimonPost: Why don't we return the error here? As said in a previous discussion about this line, our error type already supports mapping an io error to our own error.

src/net/socket.rs
```diff
@@ -159,70 +128,31 @@ impl Socket {
...
+                Ok(None)
+            }
+            Err(e) => {
+                error!("Encountered an error receiving data: {:?}", e);
```
TimonPost: Why not return the error here?

jstnlef:
Last thing is to see why the Windows builds are failing and then we can put this to bed 👍

.travis.yml
```diff
@@ -1,5 +1,9 @@
 # Set up the Rust toolchain.
 language: rust
+
+env:
+  - RUST_LOG="debug"
```
LucioFranco: you're gonna wanna set this to `RUST_LOG=laminar=debug`

fhaynes:
bors r=jstnlef,TimonPost

Issue #148, opened 2019-03-01
Implement the ordering module


TimonPost:
In #145 the arranging module was added and not yet coupled with laminar. This should be implemented for laminar.

TimonPost:
I will be doing a PR soon for implementing this.

TimonPost:
Closing this because #150

Issue #137, opened 2019-01-28
Fragment handling code doesn't invoke ExternalAcks::ack method


jstnlef:
In porting laminar to the new laminar-mio module, I was trying to get the fragmentation handling working correctly and noticed that the ExternalAcks record isn't getting updated when handling completed fragments. I tested this by updating the the following test methods in `VirtualConnection`.
`process_reliable_unordered_packet` and `process_fragmented_packet`.

```
fn process_reliable_unordered_packet() {
        let mut connection = create_virtual_connection();

        let buffer = vec![1; 500];

        for _ in 0..10 {
            let mut packet_data = connection
                .process_outgoing(&buffer, DeliveryMethod::ReliableUnordered)
                .unwrap();
            assert_eq!(packet_data.fragment_count(), 1);

            assert_packet_payload(&buffer, packet_data.parts(), &mut connection);
        }
        assert!(false);
    }
```

```
fn process_fragmented_packet() {
        let mut connection = create_virtual_connection();

        let buffer = vec![1; 4000];

        for _ in 0..10 {
            let mut packet_data = connection
                .process_outgoing(&buffer, DeliveryMethod::ReliableUnordered)
                .unwrap();

            // there should be 4 fragments
            assert_eq!(packet_data.fragment_count(), 4);

            for (index, part) in packet_data.parts().into_iter().enumerate() {
                let option = connection.process_incoming(&part).unwrap();

                // take note index 3 will contain the fragment data because the bytes of the fragmented packet will be returned when all fragments are processed.
                // that is why the last packet (index 3) can be asserted on.
                match option {
                    None => {
                        if index < 3 {
                            assert!(true)
                        } else {
                            assert!(false)
                        }
                    }
                    Some(packet) => {
                        if index == 3 {
                            assert_eq!(buffer, packet.payload());
                        }
                    }
                }
            }
        }
        assert!(false);
    }
```

I also updated `VirtualConnection::process_incoming` with the following print statement after making `their_acks` public.
```
if !ProtocolVersion::valid_version(header.protocol_version) {
            return Err(NetworkErrorKind::ProtocolVersionMismatch.into());
        }

        println!("Their acks before: {:?}", self.reliable_unordered_channel.their_acks);

        if header.packet_type_id == PacketTypeId::Fragment {
```

In the non-fragmented case, I get
```
Their acks before: ExternalAcks { last_seq: 0, field: 0, initialized: false }
[...]

Issue #96, opened 2018-11-10
Use the quality field in reliable_channel


fhaynes:
Currently we aren't using it for anything, so I've prefaced it with a _. This is a tracker to remind us to use it.

BourgondAries:
What's the intent of the field?

TimonPost:
It is originally meant for congestion management, and this enum would point out whether the network is good or bad. So in order for this value to function, we should implement that.

PR #150, opened 2019-03-08
Arranging module implementation

TimonPost:
This PR implements the [arranging](https://github.com/amethyst/laminar/tree/master/src/infrastructure/arranging) module into laminar. A view steps where necessary to accomplish this.

## Changes
In this section, I want to highlight the key points of what I changed and clarify the many changes.

1. Implemented Arranging of packets.
2. Removed Channels
3. Redesign Headers
4. Unit Tests

Those issues are solved:
#137 
#148 

### Justifying so many changes.

Don't worry, a lot of changes made are moved code, removed code, and unit tests.

1. Implemented Arranging of packets.
This was my main priority and the points below are the ones who were infected by this and needed to be updated too. 

- [Removed Channels](#Channels) (- 314 LoC)
- [Redisgning Headers](#PacketHeaders) (- 191  LoC, | + 247 LoC)
- [Refactored Packetprocessing](#PacketProcessing) 
    - Existing logic moved to own types (- 77 LoC | + 77 LoC)
    - Introduced some types to move some code out of `VirtualConnection` to  (+ 329 LoC)
     I did this to make things easier, cleaner and to prevent duplication, prevent a 1500 line code file.
    - Changes in processing code (231 LoC); this was eventually the goal of my PR, and I think I kept it pretty close to our 200 boundaries ;).
2. [Unit Tests](LINK) (+ 755 LoC): 
For the unit tests, I made 755 addictions, which is about 41% percent of the changes, which is positive right?

**Conclusion**
It seems like there are a lot of changes but to put it into perspective: 41% unit tests,  14% packet header changes, 12% PR-oriented changes, 22% were oriented around moving code, we are left with 11% which are from comments and some namespace managing.

### Channels
I removed the channels because they did not fit very well with the ordering and sequencing. The problem I encountered was that both the `ReliableChannel` and `UnrelialbeChannel` needed to access the `OrderingSystem` mutable and there were two options to do this: 
1) Mutex<Arc<<ArrangingSystem>>; this is a no-go.
2) Passing a mutable reference to the functions of those channels
This isn't ideal because I need to pass the mutable references to those functions even if those arguments are not used; e.g. Unreliable + Unordered, Reliable + Unordered, 

   Also, the trait definition would become very strange because e.g. `ReliableChannel` should eventually have access to both SequencingSystem `OrderingSystem`, and to allow that we should need to some dynamic dispatch.
3) The channel idea was a bit weird, at first I thought it would be helpful, but I noticed that it was not very much needed at all. I noticed that if I would continue with the channels that I had to do some duplication of code and that processing would eventually become more difficult. 

If those channels are removed where did the logic in them go? I moved the code of those channels to VirtualConnection; it has two methods: [process_incomming()](https://github.com/TimonPost/laminar/blob/arranging_module_implementation/src/net/connection/virtual_connection.rs#L209), [process_outgoing()](https://github.com/TimonPost/laminar/blob/arranging_module_implementation/src/net/connection/virtual_connection.rs#L81). In there I take care of processing incoming packets.

### Packet Processing
As we spoke of in [Channels](#Channel), the code is moved to VirtualConnection. We have two processes here that are of important:

1. Incoming Packet Handling
laminar_incomming

2. Outgoing Packet Handling
laminar_outgoing

3. I removed `DeliveryType` and split it up into two enums:

I did this because it would make processing easier. 
- [Ordering Guarantees](https://github.com/TimonPost/laminar/blob/arranging_module_implementation/src/packet/enums.rs#L32)
```rust
pub enum OrderingGuarantee {
    /// No arranging will be done.
    None,
    /// Packet will be arranged in sequence.
    Sequenced(Option<u8>),
    /// Packet will be arranged in order.
    Ordered(Option<u8>),
}
```
As you see `Sequenced` and `Ordered` are both containing an `Option`, this allows the user to pass in `None` as well. 
If `None` is specified then the default stream would be used; which is specified in `net::constants`. 

- [Delivery Guarantee](https://github.com/TimonPost/laminar/blob/arranging_module_implementation/src/packet/enums.rs#L5)
``` rust
pub enum DeliveryGuarantee {
    /// Packet will be delivered Unreliable
    Unreliable,
    /// Packet will be delivered Reliably
    Reliable,
}
```

**conclusion**
I've decided to put all the processing logic into `VirtualConnection` I did a view attempts to separate some processing code because there was a lot of logic, however a lot of those attempts ended up unnecessarily boilerplate code or dirty code because the processing logic has lot's of common code and spreading this only increased the sharing of variables and such.

### Packet Headers
For arranging, I had to modify the packet headers a bit. 

1) I've introduced a new header called [ArrangingHeader](https://github.com/TimonPost/laminar/blob/arranging_module_implementation/src/packet/header/arranging_header.rs), this header contains two values: the `order_id` and the `stream_id`, those are needed to know at which 'stream' the ordering should be done specified on the given 'id'.
2) I removed the dependencies of headers to each other; `AcknowledgmentHeader` had a `StandardHeader`, `FragmentHeader` had an `AcknowledgmentHeader`. Until now this was valid, but, the new `ArrangingHeader` could be constructed with another header in two different ways: 
- `StandardHeader` -> `AcknowledgmentHeader` -> `ArrangingHeader`; in case of 'Reliable + Ordered'
- `StandardHeader` -> `ArrangingHeader`; in case of 'Unreliable + Sequenced'

I needed to make each header to be a totally independent header and to have a static size because the headers of a packet vary on the `DeliveryGuarantee` and `OrderingGuarantee`. 

Let's take a look at this diagram first.

_Header Diagram_
New Project (1)

As you see each packet type has its own header, we need to do some parsing - as we did before - to reconstruct the given headers when receiving a packet.

The constructing and reconstructing of packets and headers is the main part why this PR is a little too big. I introduced two types to abstract away a bit of the processing code, and to prevent to much code in `VirtualConnection`:

- [PacketReader](https://github.com/TimonPost/laminar/blob/arranging_module_implementation/src/packet/packet_reader.rs)
Could be used to read the packet contents of laminar, where 'contents' is the headers and payload.
- [OutgoingPacketBuilder](https://github.com/TimonPost/laminar/blob/arranging_module_implementation/src/packet/outgoing.rs#L12)
A builder that could be used to construct an outgoing laminar packet, the packet includes the header and payload. 
- [Outgoing](https://github.com/TimonPost/laminar/blob/arranging_module_implementation/src/packet/outgoing.rs#L126)
 Enum for storing different kinds of outgoing types with data.
```
pub enum Outgoing<'a> {
    /// Represents a single packet.
    Packet(OutgoingPacket<'a>),
    /// Represents a packet that is fragmented and thus contains more than one `OutgoingPacket`.
    Fragments(Vec<OutgoingPacket<'a>>),
}
```
- [OutgoingPacket](https://github.com/TimonPost/laminar/blob/arranging_module_implementation/src/packet/outgoing.rs#L107)
```
pub struct OutgoingPacket<'p> {
    header: Vec<u8>,
    payload: &'p [u8],
}
```

As you might already see I introduced a packet which has lifetimes to some buffer. 
The idea behind this is, is that the data the user sends into laminar could be a reference until it is sent on the socket. The header could not be a slice, because those are dynamic and dependent on the packet type.

## Unit Tests
The most changes about 41% were caused by new unit tests. To assure that my code is working, I have tests for packet order, different reliabilities, packet serialization, deserialization, etc. I could do some more but the PR-size held me back I will do some later.

# Conclusion

I know the size of this PR is big, I don't mind if some time has to be taken to review it. Although as I noticed before, not all changes are new code and not all file changes are real but most of them are caused by the rename `DeliveryType` to `DeliveryGuarantee`.


src/error/network_error.rs
```diff
@@ -28,6 +30,8 @@ pub enum ErrorKind {
     JoiningThreadFailed,
     /// There was an unexpected error caused by an poisoned lock.
```
fhaynes:
```suggestion
    /// There was an unexpected error caused by a poisoned lock.
```

src/infrastructure/acknowlegement.rs
```diff
@@ -0,0 +1,42 @@
...
+        AcknowledgementHandler {
+            seq_num: 0,
+            waiting_packets: Default::default(),
+            their_acks: Default::default(),
+            dropped_packets: Vec::new(),
+        }
+    }
+}
+
+impl AcknowledgementHandler {
```
fhaynes: Can we get /// comments for all the pub functions, please?

src/infrastructure/acknowlegement.rs
```diff
@@ -0,0 +1,42 @@
+use crate::net::{ExternalAcks, LocalAckRecord};
+
+/// Type responsible for handling the acknowledgement of packets.
+pub struct AcknowledgementHandler {
+    waiting_packets: LocalAckRecord,
+    their_acks: ExternalAcks,
+    pub seq_num: u16,
```
fhaynes: At some point, it might be worth aliasing u16 to like SequenceNumber. We use it in a lot of places, and it might be clearer than just u16.

src/infrastructure/arranging/ordering.rs
```diff
@@ -171,6 +174,12 @@ impl<T> OrderingStream<T> {
         self.expected_index
     }
 
+    /// Returns the unique identifier which should be used for ordering on an other stream e.g. the remote endpoint.
```
fhaynes:
```suggestion
    /// Returns the unique identifier which should be used for ordering on the other stream e.g. the remote endpoint.
```

src/infrastructure/arranging/sequencing.rs
```diff
@@ -10,10 +10,7 @@
 //! - See [super-module](../index.html) description for more details.
 
 use super::{Arranging, ArrangingSystem};
-use std::{
-    collections::HashMap,
-    marker::PhantomData
-};
+use std::{collections::HashMap, marker::PhantomData};
 
 /// An sequencing system that can arrange items in sequence on different streams.
```
fhaynes:
```suggestion
/// A sequencing system that can arrange items in sequence across different streams.
```

src/infrastructure/congestion.rs
```diff
@@ -0,0 +1,36 @@
...
+};
+
+use std::time::Instant;
+
+/// Type responsible for keeping track of congestion information.
+pub struct CongestionHandler {
+    rtt_measurer: RttMeasurer,
+    rtt: f32,
+    congestion_data: SequenceBuffer<CongestionData>,
+    _quality: NetworkQuality,
```
fhaynes: Why the _? Are we not using it anywhere?

src/infrastructure/congestion.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub fn new(config: &Config) -> CongestionHandler {
+        CongestionHandler {
+            rtt_measurer: RttMeasurer::new(config),
+            congestion_data: SequenceBuffer::with_capacity(<u16>::max_value() as usize),
+            _quality: NetworkQuality::Good,
+            rtt: 0.0,
+        }
+    }
+
+    pub fn process_incoming(&mut self, incoming_seq: u16) {
```
fhaynes: Comment please

src/infrastructure/congestion.rs
```diff
@@ -0,0 +1,36 @@
...
+        let congestion_data = self.congestion_data.get_mut(incoming_seq);
+        self.rtt = self.rtt_measurer.get_rtt(congestion_data);
+    }
+
+    pub fn process_outgoing(&mut self, seq: u16) {
```
fhaynes: Comment please

src/infrastructure/fragmenter.rs
```diff
@@ -50,7 +51,7 @@ impl Fragmentation {
     ///
     /// So for 4000 bytes we need 4 fragments
     /// [fragment: 1024] [fragment: 1024] [fragment: 1024] [fragment: 928]
-    pub fn total_fragments_needed(payload_length: u16, fragment_size: u16) -> u16 {
+    pub fn fragments_needed(payload_length: u16, fragment_size: u16) -> u16 {
         let remainder = if payload_length % fragment_size > 0 {
             1
```
fhaynes: Should this be a constant for clarity?

src/infrastructure/fragmenter.rs
```diff
@@ -60,30 +61,21 @@ impl Fragmentation {
...
+        payload: &'a [u8],
         config: &Config,
-    ) -> Result<()> {
+    ) -> Result<Vec<&'a [u8]>> {
+        let mut fragments = Vec::new();
+
         let payload_length = payload.len() as u16;
         let num_fragments =
-            Fragmentation::total_fragments_needed(payload_length, config.fragment_size) as u8; /* safe cast max fragments is u8 */
+            Fragmentation::fragments_needed(payload_length, config.fragment_size) as u8; /* safe cast max fragments is u8 */
```
fhaynes: Can we put the comment above the line just to keep the style consistent? I'm also not totally sure what this comment means.

```suggestion
            // Safe cast max fragments is u8
            Fragmentation::fragments_needed(payload_length, config.fragment_size) as u8; 
```

src/infrastructure/fragmenter.rs
```diff
@@ -60,30 +61,21 @@ impl Fragmentation {
...
-            let fragment = FragmentHeader::new(
-                acked_header.standard_header,
-                fragment_id,
-                num_fragments,
-                acked_header,
-            );
-            let mut buffer = Vec::with_capacity(fragment.size() as usize);
-            fragment.parse(&mut buffer)?;
-
             // get start end pos in buffer
```
fhaynes:
```suggestion
            // get start and end position of buffer
```

src/infrastructure/fragmenter.rs
```diff
@@ -159,28 +150,17 @@ impl Fragmentation {
...
-                    None => Err(FragmentErrorKind::PacketHeaderNotFound)?,
-                }
-            } else {
-                Err(FragmentErrorKind::AlreadyProcessedFragment)?
-            }
-        }
+            let reassembly_data = ReassemblyData::new(
+                fragment_header.sequence(),
+                fragment_header.fragment_count(),
+                (9 + self.config.fragment_size) as usize
```
fhaynes: Can we make the 9 a constant, or have a comment explaining it?

src/net/socket.rs
```diff
@@ -131,34 +123,52 @@ impl Socket {
...
             }
+
+            return Ok(bytes_sent);
         }
 
-        Ok(bytes_sent)
+        Ok(0)
     }
 
     // Receives a single message from the socket. On success, returns the packet containing origin and data.
```
fhaynes: Since this isn't returning the packet anymore, we should probably update the comment

src/net/socket.rs
```diff
@@ -131,34 +123,52 @@ impl Socket {
...
-    fn recv_from(&mut self) -> Result<Option<Packet>> {
+    fn recv_from(&mut self) -> Result<()> {
         let (recv_len, address) = self.socket.recv_from(&mut self.recv_buffer)?;
         if recv_len == 0 {
```
fhaynes: Possibly make the 0 a constant just for clarity

src/packet/enums.rs
```diff
@@ -0,0 +1,137 @@
+use crate::packet::EnumConverter;
+
+/// Enum to specify how a packet should be delivered.
+#[derive(Copy, Clone, Debug, PartialOrd, PartialEq, Eq)]
+pub enum DeliveryGuarantee {
+    /// Packet will be delivered Unreliable
```
fhaynes:
```suggestion
    /// Packet will be delivered unreliably
```

src/packet/enums.rs
```diff
@@ -0,0 +1,137 @@
...
+    Unreliable,
+    /// Packet will be delivered Reliably
```
fhaynes:
```suggestion
    /// Packet will be delivered reliably
```

src/packet/header/arranging_header.rs
```diff
@@ -0,0 +1,99 @@
+use super::{HeaderReader, HeaderWriter};
+use crate::error::Result;
+use crate::net::constants::ARRANGING_PACKET_HEADER;
+use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
+use std::io::Cursor;
+
+#[derive(Copy, Clone, Debug)]
+/// This header represents a fragmented packet header.
+pub struct ArrangingHeader {
+    arranging_id: u16,
```
fhaynes: yeah I definitely think some type aliases are in order

src/packet/header/standard_header.rs
```diff
@@ -1,37 +1,70 @@
...
             protocol_version: ProtocolVersion::get_crc32(),
-            packet_type_id,
-            delivery_method,
+            packet_type,
+            delivery_guarantee,
+            ordering_guarantee,
         }
     }
+
+    pub fn protocol_version(&self) -> u32 {
```
fhaynes:
```suggestion
    /// Returns the protocol version
    pub fn protocol_version(&self) -> u32 {
```

src/packet/header/standard_header.rs
```diff
@@ -1,37 +1,70 @@
...
+        self.protocol_version
+    }
+
+    pub fn delivery_guarantee(&self) -> DeliveryGuarantee {
```
fhaynes:
```suggestion
    /// Returns the DeliveryGuarantee
    pub fn delivery_guarantee(&self) -> DeliveryGuarantee {
```

src/packet/header/standard_header.rs
```diff
@@ -1,37 +1,70 @@
...
+        self.delivery_guarantee
+    }
+
+    pub fn ordering_guarantee(&self) -> OrderingGuarantee {
```
fhaynes:
```suggestion
    /// Returns the OrderingGuarantee
    pub fn ordering_guarantee(&self) -> OrderingGuarantee {
```

src/packet/header/standard_header.rs
```diff
@@ -1,37 +1,70 @@
...
+        self.ordering_guarantee
+    }
+
+    pub fn packet_type(&self) -> PacketType {
```
fhaynes:
```suggestion
    /// Returns the PacketType
    pub fn packet_type(&self) -> PacketType {
```

src/packet/header/standard_header.rs
```diff
@@ -1,37 +1,70 @@
...
+        self.packet_type
+    }
+
+    pub fn is_fragment(&self) -> bool {
```
fhaynes:
```suggestion
    /// Returns true if the packet is a fragment, false if not
    pub fn is_fragment(&self) -> bool {
```

src/packet/header/standard_header.rs
```diff
@@ -1,37 +1,70 @@
...
+        self.packet_type == PacketType::Fragment
+    }
+
+    pub fn is_current_protocol(&self) -> bool {
```
fhaynes:
```suggestion
    /// Checks if the protocol version in the packet is a valid version
    pub fn is_current_protocol(&self) -> bool {
```

src/packet/header/standard_header.rs
```diff
@@ -72,28 +108,51 @@ impl HeaderReader for StandardHeader {
...
+    fn serialize() {
+        let mut buffer = Vec::new();
+        let header = StandardHeader::new(
+            DeliveryGuarantee::Unreliable,
+            OrderingGuarantee::Sequenced(None),
+            PacketType::Packet,
+        );
+        header.parse(&mut buffer).is_ok();
+
+        // [0 .. 3] protocol version
```
fhaynes: Not sure what this comment means.

src/packet/outgoing.rs
```diff
@@ -0,0 +1,202 @@
...
+    },
+};
+
+/// Builder that could be used to construct an outgoing laminar packet.
+pub struct OutgoingPacketBuilder<'p> {
+    header: Vec<u8>,
+    payload: &'p [u8],
+}
+
+impl<'p> OutgoingPacketBuilder<'p> {
```
fhaynes: Should we be using expect in this section, since it will panic? Why not return an Result?

src/infrastructure/acknowlegement.rs
TimonPost: I see why you want that, I left it away because it is not public for the user.

src/infrastructure/acknowlegement.rs
TimonPost: Yea, good one.

src/infrastructure/congestion.rs
TimonPost: We already created this issue: #96

src/infrastructure/congestion.rs
TimonPost: I see why you want that, I left it away because it is not public for the user.

src/infrastructure/congestion.rs
TimonPost: I see why you want that, I left it away because it is not public for the user.

src/infrastructure/fragmenter.rs
TimonPost: Nah, because it means just one fragment. Don't see the reason to move it into a constant.

src/infrastructure/fragmenter.rs
TimonPost: Yea, probably better.

src/net/socket.rs
TimonPost: You mean: `ZERO = 0` ?  Do we really need constants for those numbers?

src/net/events.rs
```diff
@@ -1,11 +1,13 @@
...
 use std::net::SocketAddr;
 
 /// Events which will be pushed through the event_receiver returned by `Socket::bind`.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub enum SocketEvent {
     /// A packet has been received from a client.
     Packet(Packet),
+    /// An fragmentation has completed and this contains the completed data.
+    Fragment(Fragment),
```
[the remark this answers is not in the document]
TimonPost: `Fragment` contains some other data because it doesn't have `OrderingGuarantee` or `DeliveryGuarantee`. It only has the addr and payload. So it is a trade-off between creating a `Packet` with `OrderingGuatantee` being not specified or put to `None` or leaving this here. The second reason why I did this to allow the user to handle fragments differently if they want to.

src/packet/outgoing.rs
TimonPost: I did this for two reasons:

1.  When using a builder and returning `Self` we can't use the API properly because you need  to unwrap:

```
buidler.with_fragment_header()?.with_default_header()?.with_sequencing_header()?;
````

2. If the parsing of our headers is wrong, I don't see any reason to continue with receiving the packet. We can choose to stop handling this packet. However, parsing code and invalid headers are probably an issue of our code and it'll probably be failing for the next packet too. It could be possible that send data is invalid, but we have most of this covered with the CRC32, although one can construct a packet with the right CRC32 and create its own packet.

Now I think of this we should probably have a CRC32 over the whole packet content too.

File: src/packet/header/standard_header.rs (hunk 24 above)
Comment (TimonPost) [reply src/packet/header/standard_header.rs]: I skipped the first 4 bytes who are the CRC32, however, I can't easily hardcode those into the `vec!` above it, simply because I don't know the CRC32 bytes yet. So I decided to just skip those bytes and start asserting on the other bytes.

File: src/packet/header/arranging_header.rs (hunk 17 above)
Comment (TimonPost) [reply src/packet/header/arranging_header.rs]: Right, let me do that for you.

File: src/net/events.rs (hunk 26 above)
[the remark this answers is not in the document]
Comment (jstnlef) [reply src/net/events.rs]: So I ask you this question. When using TCP, do you have to (or want to) distinguish between a message that was fragmented or not?

File: src/packet/outgoing.rs (hunk 25 above)
Comment (fhaynes) [reply src/packet/outgoing.rs]: TCP handles its own fragmentation. We shouldn't be adding anything on top of TCP, I would think...

File: src/net/events.rs (hunk 26 above)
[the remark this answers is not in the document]
Comment (fhaynes) [reply src/net/events.rs]: TCP handles its own fragmentation. We shouldn't be adding anything on top of TCP, I would think...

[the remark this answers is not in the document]
Comment (TimonPost) [reply src/net/events.rs]: @fhaynes No that is not the question here, @jstnlef yea, I thought of that too, alright let's `Packet` only.

**Comment by jstnlef** [comment]
One other piece of advice, from me at least, is to not let PR size hold you back from writing unit tests. I much prefer a 2000+ line PR where 1500+ lines are unit tests than to not include them because of some mythical "this PR is too big" number

File: src/packet/outgoing.rs (hunk 25 above)
Comment (jstnlef) [reply src/packet/outgoing.rs]: The CRC32 should definitely include the payload. Otherwise, it isn't really worth having to be honest.

**Comment by TimonPost** [comment]
Alright, can I agree on that, but at least the code I changed is covered by unittests, so when this is merged we or I can create other unit tests.

File: src/infrastructure/acknowlegement.rs (hunk 2 above)
Comment (TimonPost) [reply src/infrastructure/acknowlegement.rs]: done

File: src/infrastructure/congestion.rs (hunk 8 above)
Comment (TimonPost) [reply src/infrastructure/congestion.rs]: done

File: src/infrastructure/congestion.rs (hunk 7 above)
Comment (TimonPost) [reply src/infrastructure/congestion.rs]: done

File: src/infrastructure/fragmenter.rs (hunk 9 above)
Comment (TimonPost) [reply src/infrastructure/fragmenter.rs]: I am closing this because I think it is unnecessarily.

File: src/packet/header/arranging_header.rs (hunk 17 above)
Comment (TimonPost) [reply src/packet/header/arranging_header.rs]: done

File: src/packet/outgoing.rs (hunk 25 above)
Comment (TimonPost) [reply src/packet/outgoing.rs]: Yea, the idea is that the CRC32 is smaller than a string for a header. We need to check on the protocol version, I think that CRC32 is fitting this purpose well. 

However, we probably want to have a CRC32 over the whole payload. I will create an issue for this.

Comment (TimonPost) [reply src/packet/outgoing.rs]: https://github.com/amethyst/laminar/issues/151

File: src/packet/outgoing.rs [hunk 27]
```diff
@@ -0,0 +1,202 @@
...
+    packet::{
+        header::{
+            AckedPacketHeader, ArrangingHeader, FragmentHeader, HeaderWriter, StandardHeader,
+        },
+        DeliveryGuarantee, OrderingGuarantee, PacketType,
+    },
+};
+
+/// Builder that could be used to construct an outgoing laminar packet.
+pub struct OutgoingPacketBuilder<'p> {
```
jstnlef: For this builder, I think it makes sense to store all of the intermediate bits on the builder and only parse to the Outgoing packet on build. This allows us to only have one allocation of a Vec when the outgoing packet is built.
This would make the signature of the `OutgoingPacket` be 
```
struct OutgoingPacket{
    contents: Vec<u8>
}
```

src/packet/packet_reader.rs
```diff
@@ -0,0 +1,127 @@
...
+    pub fn read_standard_header(&mut self) -> Result<StandardHeader> {
+        self.cursor.set_position(0);
+        StandardHeader::read(&mut self.cursor)
+    }
+
+    /// Read the `StandardHeader` from the underlying buffer.
+    ///
+    /// # Remark
+    /// - Will change the position to the location of `StandardHeader`
+    pub fn read_arranging_header(&mut self, start_offset: u16) -> Result<ArrangingHeader> {
```
jstnlef: I worry that these methods don't have a check to see if it's possible that this particular buffer should have an `ArrangingHeader` before return an `ArrangingHeader` with junk bytes it read off of the slice.

src/packet/packet_reader.rs
```diff
@@ -0,0 +1,127 @@
...
+        self.cursor.set_position(u64::from(start_offset));
+        ArrangingHeader::read(&mut self.cursor)
+    }
+
+    /// Read the `AckedPacketHeader` from the underlying buffer.
+    ///
+    /// # Remark
+    /// - Will change the position to the location of `AckedPacketHeader`
+    pub fn read_acknowledge_header(&mut self) -> Result<AckedPacketHeader> {
```
jstnlef: Same here.

src/packet/packet_reader.rs
```diff
@@ -0,0 +1,127 @@
...
+        Ok((fragment_header, acked_header))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::packet::header::HeaderReader;
+    use crate::packet::{DeliveryGuarantee, OrderingGuarantee, PacketReader, PacketType};
+
+    #[test]
```
jstnlef: I would like to see quite a few more tests here.

src/packet/packet_structure.rs
```diff
@@ -1,28 +1,48 @@
...
     /// the raw payload of the packet
     payload: Box<[u8]>,
     /// defines on how the packet will be delivered.
-    delivery_method: DeliveryMethod,
+    delivery: DeliveryGuarantee,
+    /// defines on how the packet will be ordered.
+    ordering: OrderingGuarantee,
 }
 
 impl Packet {
```
jstnlef: Looks like you're missing `reliable_sequenced` and subsequent tests

src/packet/header/standard_header.rs
```diff
@@ -1,37 +1,78 @@
...
+    }
+
+    /// Returns true if the packet is a fragment, false if not
+    pub fn is_fragment(&self) -> bool {
+        self.packet_type == PacketType::Fragment
+    }
+
+    /// Checks if the protocol version in the packet is a valid version
+    pub fn is_current_protocol(&self) -> bool {
+        if !ProtocolVersion::valid_version(self.protocol_version) {
```
jstnlef: Big pet peeve of mine. This should just be `ProtocolVersion::valid_version(self.protocol_version)`

src/lib.rs
```diff
@@ -37,9 +37,19 @@ mod net;
...
-pub use self::infrastructure::DeliveryMethod;
 pub use self::net::Socket;
 pub use self::net::SocketEvent;
+pub use self::packet::{DeliveryGuarantee, OrderingGuarantee, Packet};
 pub use self::net::VirtualConnection;
-pub use self::packet::Packet;
 pub use self::protocol_version::ProtocolVersion;
+
+/// Interface for converting a type into a boxed slice.
+trait IntoBoxedSlice {
```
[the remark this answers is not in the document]
TimonPost: Yea, you are right, I introduced this because I was using it at more places, but it can be removed now.

src/net/connection/virtual_connection.rs
```diff
@@ -1,51 +1,55 @@
...
-            unreliable_unordered_channel: UnreliableChannel::new(true),
-            reliable_unordered_channel: ReliableChannel::new(false, config),
-            sequenced_channel: SequencedChannel::new(),
-
+            ordering_system: OrderingSystem::new(),
+            sequencing_system: SequencingSystem::new(),
+            acknowledge_handler: AcknowledgementHandler::new(),
+            congestion_handler: CongestionHandler::new(config),
             fragmentation: Fragmentation::new(config),
+            config: config.clone(),
```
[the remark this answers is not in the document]
TimonPost: Yes, it does make sense and I thought about that also, I wanted to move that to another PR, but as far this PR concerns I could remove this clone here and just make this VirtualConnection own it.

src/packet/outgoing.rs
TimonPost: I am not sure what you mean by ' intermediate bits'. I already have a reference to the payload in here. And it does already do one allocation over [here](https://github.com/amethyst/laminar/pull/150/files#diff-71b90166b67cde207975f7515922c988R118), when it is build.

src/packet/packet_reader.rs
TimonPost: Right, let me make some validation methods over there.

src/packet/packet_reader.rs
TimonPost: Right, let me make some validation methods over there.

src/packet/header/standard_header.rs
TimonPost: This is actually really stupid of me, lol, thanks for highlighting this.

src/packet/outgoing.rs
jstnlef: As you're adding bytes to the header `Vec`, the vector needs to grow (since you didn't initialize it with a capacity). Each time this vector grows (roughly doubling in size) it allocates a new underlying section of memory. Then, when you call `concat()` it does yet another allocation into a new `Vec` and returns it. My suggestion is to store all of the stack structures (the header structs) on the builder as defined, and only populate a new vector when the call to `build()` is made.

src/packet/packet_reader.rs
jstnlef: Think of it this way, if I send you a bunch of garbage bytes over the socket, I would want the reader to error if the buffer doesn't represent a properly formatted laminar packet. I'm less worried about the case in which we run out of bytes to read. More that those bytes may not be what we expect and we should protect against that at the reader level.

src/packet/packet_structure.rs
TimonPost: I left this one away due to some troubles before, I revisited this guarantee and I think I could make it work.

src/infrastructure/arranging/ordering.rs
```diff
@@ -128,12 +129,14 @@ impl<'a, T> ArrangingSystem for OrderingSystem<T> {
 /// - See [super-module](../index.html) for more information about streams.
 pub struct OrderingStream<T> {
     // the id of this stream.
-    stream_id: u8,
+    _stream_id: u8,
```
fhaynes: Is this for future use?

src/infrastructure/arranging/ordering.rs
```diff
@@ -128,12 +129,14 @@ impl<'a, T> ArrangingSystem for OrderingSystem<T> {
...
     // the storage for items that are waiting for older items to arrive.
     // the items will be stored by key and value where the key is de incoming index and value the item value.
```
fhaynes:
```suggestion
    // the items will be stored by key and value where the key is the incoming index and the value is the item value.
```

src/infrastructure/fragmenter.rs
```diff
@@ -60,31 +61,20 @@ impl Fragmentation {
...
-        packet_data: &mut PacketData,
-        config: &Config,
-    ) -> Result<()> {
+    pub fn spit_into_fragments<'a>(payload: &'a [u8], config: &Config) -> Result<Vec<&'a [u8]>> {
+        let mut fragments = Vec::new();
+
         let payload_length = payload.len() as u16;
         let num_fragments =
-            Fragmentation::total_fragments_needed(payload_length, config.fragment_size) as u8; /* safe cast max fragments is u8 */
+            // Safe cast max fragments is u8
```
fhaynes: I don't quite understand what this comment is trying to say

src/infrastructure/fragmenter.rs
```diff
@@ -159,28 +148,17 @@ impl Fragmentation {
...
-                    None => Err(FragmentErrorKind::PacketHeaderNotFound)?,
-                }
-            } else {
-                Err(FragmentErrorKind::AlreadyProcessedFragment)?
-            }
-        }
+            let reassembly_data = ReassemblyData::new(
+                fragment_header.sequence(),
+                fragment_header.fragment_count(),
+                (9 + self.config.fragment_size) as usize,
```
fhaynes: Probably a comment on what the 9 is for, or using a constant, would help clarity here

src/lib.rs
```diff
@@ -22,7 +22,7 @@
 //!
 
 #![warn(missing_docs)]
-
+#![allow(clippy::trivially_copy_pass_by_ref)]
```
fhaynes: What do we need this for?

src/packet/header/arranging_header.rs
```diff
@@ -0,0 +1,99 @@
...
+        header.parse(&mut buffer).is_ok();
+
+        assert_eq!(buffer[1], 1);
+        assert_eq!(buffer[2], 2);
+    }
+
+    #[test]
+    fn deserialize() {
+        let buffer = vec![0, 1, 2];
+
```
fhaynes:
```suggestion
```

src/packet/header/arranging_header.rs
```diff
@@ -0,0 +1,99 @@
...
+        let mut cursor = Cursor::new(buffer.as_slice());
+
```
fhaynes:
```suggestion
```

src/packet/packet_reader.rs
fhaynes: This should probably have an effect on connection quality if enough happen, along with the arranging header.

fhaynes:
> One other piece of advice, from me at least, is to not let PR size hold you back from writing unit tests. I much prefer a 2000+ line PR where 1500+ lines are unit tests than to not include them because of some mythical "this PR is too big" number

It takes on average 30 minutes to review 200 lines of code. If your team's policy is to review tests with the same rigor as code changes, it should. There is definitely a "this PR is too big" number, though. Human attention span being what it is, anything that takes > 20-30 minutes tends to put people into "skim" mode, and the number of bugs and regressions jump.

TimonPost:
This PR is about 560 LoC issue realated. Due to more added unit tests this PR is now over 50% unit tests. The other changed code is about moving existing logic, which doesn't need to have a review.

@jsfnl was fine with lots of tests and would like to see 2000 lines PR with 1500 lines of unit tests instead of 2000 lines PR with 0. You said that it was "ok" to have lots of tests.

I did my best to keep it small, but if there was a limit of 200 lines it was impossible to add this feature. 200 lines is really small. I can see that this works out in a more mature library. I think 550 LoC is acceptabel. If tests where also includes I would have left them away which is definitely not plausible.  They at least assure that those changes are right.

src/infrastructure/fragmenter.rs
TimonPost: Not related to this PR, but it means that the maximal fragments are restricted to 256; u8. And thus this cast is safe.

src/lib.rs
TimonPost: We discussed this in discord, Clippy showed '& self' in a function as an error and suggested me to change it too 'self' which I don't want because it would consume self. I added this lint to prevent the CI from braking.

jstnlef:
Looks like the tests are still broken.

src/infrastructure/congestion.rs
```diff
@@ -0,0 +1,44 @@
...
+    sequence_buffer::{CongestionData, SequenceBuffer},
+    Config,
+};
+
+use std::time::Instant;
+
+/// Type that is responsible for keeping track of congestion information.
+pub struct CongestionHandler {
+    rtt_measurer: RttMeasurer,
+    rtt: f32,
```
jstnlef: Probably makes sense to have `rtt` be a method (`rtt()`) where the calculated value of `rtt` lives in the `RttMeasurer`

src/infrastructure/fragmenter.rs
jstnlef: Agreed. This should be a constant.

src/net/connection/virtual_connection.rs
```diff
@@ -56,87 +60,345 @@ impl VirtualConnection {
     }
 
     /// This pre-process the given buffer to be send over the network.
-    /// 1. It will append the right header.
-    /// 2. It will perform some actions related to how the packet should be delivered.
-    pub fn process_outgoing(
+    ///
+    /// # Algorithm
+    /// 1. Write `StandardHeader`
```
jstnlef: As nice as these are today, I really dislike having the algorithm description as a comment. Inevitably, we're going to make changes to the code and forget to update the comment which will leave the comment misleading/useless. Let the code speak for itself please :).

src/net/connection/virtual_connection.rs
```diff
@@ -56,87 +60,345 @@ impl VirtualConnection {
...
 
-    /// This process the incoming data and returns an packet if the data is complete.
+    /// This processes the incoming data and returns an packet if the data is complete.
     ///
-    /// Returns `Ok(None)`:
-    /// 1. In the case of fragmentation and not all fragments are received
-    /// 2. In the case of the packet being queued for ordering and we are waiting on older packets first.
-    pub fn process_incoming(&mut self, received_data: &[u8]) -> Result<Option<Packet>> {
+    /// # Algorithm
+    /// 1. Read `StandardHeader`
```
jstnlef: Same here.

src/infrastructure/arranging/ordering.rs
TimonPost: Yea, every stream has an id, this value is not used in the stream itself, but I like to have it there to notice that there is a stream-id bound to it.

src/bin/laminar-tester.rs
```diff
@@ -167,11 +167,7 @@ fn run_client(config: ClientConfiguration) -> Result<()> {
 fn test_steady_stream(sender: &Sender<Packet>, config: ClientConfiguration) {
     info!("Beginning steady-state test");
 
-    let test_packet = Packet::new(
-        config.listen_host,
-        config.test_name.into_bytes().into_boxed_slice(),
-        DeliveryMethod::ReliableUnordered,
-    );
+    let test_packet = Packet::unreliable(config.listen_host, config.test_name.into_bytes());
```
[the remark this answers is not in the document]
TimonPost: Right, let me change that.

src/infrastructure/congestion.rs
TimonPost: good one!

src/net/connection/virtual_connection.rs
TimonPost: That is part of the problem, I agree that the code should speak for its self. But I walk some problems here and I like your thoughts on it. 

I was thinking about moving all the processing logic into different types and have some extensible way to handle lot's of different parsing scenario's, but I came to the conclusion that putting the processing logic in a single function was making

There where two reasons why I did this:
1. We don't have a lot of handling scenarios and thus we don't need a very extensible way of handling packets as described in #143.  
2.  It is much easier to handle an incoming or outgoing packet if the processing logic is at one place. 

One downside that comes from processing packets in one function is that the function will be a bit bigger than normal functions should be. One can state that they only might be around 7 lines because that is the attention span of the reader. So I thought that it would be helpful for the user to add a short bullet list to help the reader to understand what this does. 

Three questions and notices:

1. Do you think we should break this down and make this packet processing more extensible?
2. Do you think this code does not explain itself? 
3.   Do I need to remove those comments; I can do that but I don't really see the purpose of that since I think it is valuable for the reader. But sure if you want to I can remove it. 

I was already thinking of braking this more down, and see what more I can do to make it a bit smaller but keep the easy maintainability of the code in a next PR because this PR would become to big.

src/net/connection/virtual_connection.rs
TimonPost: see other comment

src/infrastructure/arranging.rs
```diff
@@ -39,8 +39,11 @@
...
 
-pub mod ordering;
-pub mod sequencing;
+mod ordering;
+mod sequencing;
+
+pub use self::ordering::{IterMut, OrderingStream, OrderingSystem};
+pub use self::sequencing::{SequencingStream, SequencingSystem};
 
 /// A trait which could be implemented for arranging operations.
```
fhaynes:
```suggestion
/// A trait which that can be implemented for arranging operations.
```

src/infrastructure/arranging/ordering.rs
```diff
@@ -128,12 +129,14 @@ impl<'a, T> ArrangingSystem for OrderingSystem<T> {
...
     // the id of this stream.
-    stream_id: u8,
+    _stream_id: u8,
     // the storage for items that are waiting for older items to arrive.
-    // the items will be stored by key and value where the key is de incoming index and value the item value.
+    // the items will be stored by key and value where the key is the incoming index and the value is the item value.
     storage: HashMap<usize, T>,
     // the next expected item index.
     expected_index: usize,
+    // unique identifier which should be used for ordering on an other stream e.g. the remote endpoint.
```
fhaynes:
```suggestion
    // unique identifier which should be used for ordering on a different stream e.g. the remote endpoint.
```

src/net/connection/virtual_connection.rs
jstnlef: 7 lines!? That's extremely small. While methods/functions should err on the smaller side, having a function that is 50-100 lines long is common and reasonable in my opinion. The heuristic I like to use is whether or not I can see the entire function's implementation without scrolling. Any bigger and I consider breaking it up a little. There is such a thing as having too much abstraction. It makes the code hard to read in the other direction because you're constantly jumping between methods/modules/files to get a full picture of what's going on. So to specifically answer your questions,
1. I think this is perfectly fine
2. I think the code explains itself pretty well.
3. I would remove the comments explaining the algorithm because the code itself is explaining the algorithm in a more accurate way.

src/infrastructure/fragmenter.rs
```diff
@@ -96,18 +86,21 @@ impl Fragmentation {
...
 
-            packet_data.add_fragment(&buffer, fragment_data)?;
+            fragments.push(fragment_data);
         }
 
-        Ok(())
+        Ok(fragments)
     }
 
     /// This will read fragment data and returns the complete packet data when all fragments are received.
```
fhaynes:
```suggestion
    /// This will read fragment data and return the complete packet when all fragments are received.
```

src/packet/enums.rs
```diff
@@ -0,0 +1,137 @@
+use crate::packet::EnumConverter;
+
+/// Enum to specify how a packet should be delivered.
+#[derive(Copy, Clone, Debug, PartialOrd, PartialEq, Eq)]
+pub enum DeliveryGuarantee {
+    /// Packet will be delivered unreliably
```
fhaynes:
```suggestion
    /// Packet may or may not be delivered
```

src/packet/enums.rs
```diff
@@ -0,0 +1,137 @@
...
+    Unreliable,
+    /// Packet will be delivered reliably
```
fhaynes:
```suggestion
    /// Packet will be delivered
```

src/packet/enums.rs
```diff
@@ -0,0 +1,137 @@
...
+        }
+    }
+}
+
+/// Enum to specify how a packet should be arranged.
+#[derive(Copy, Clone, Debug, PartialOrd, PartialEq, Eq)]
+pub enum OrderingGuarantee {
+    /// No arranging will be done.
+    None,
+    /// Packet will be arranged in sequence.
```
fhaynes:
```suggestion
    /// Packets will be arranged in sequence.
```

src/packet/enums.rs
```diff
@@ -0,0 +1,137 @@
...
+    Sequenced(Option<u8>),
+    /// Packet will be arranged in order.
```
fhaynes:
```suggestion
    /// Packets will be arranged in order.
```

src/net/connection/virtual_connection.rs
TimonPost: Alright, nice to hear, let me remove them.

src/packet/packet_reader.rs
```diff
@@ -0,0 +1,249 @@
+use crate::net::constants::STANDARD_HEADER_SIZE;
+use crate::packet::header::{
+    AckedPacketHeader, ArrangingHeader, FragmentHeader, HeaderReader, StandardHeader,
+};
+use crate::{ErrorKind, Result};
+
+use std::io::Cursor;
+
+/// Could be used to read the packet contents of laminar.
```
fhaynes:
```suggestion
/// Can be used to read the packet contents of laminar.
```

src/packet/packet_reader.rs
```diff
@@ -0,0 +1,249 @@
...
+        if self.can_read(AckedPacketHeader::size()) {
+            AckedPacketHeader::read(&mut self.cursor)
+        } else {
+            Err(ErrorKind::CouldNotReadHeader(String::from(
+                "acknowledgement",
+            )))
+        }
+    }
+
+    /// Read the `FragmentHeader` and optional the `AckedPacketHeader` from the underlying buffer.
```
fhaynes:
```suggestion
    /// Read the `FragmentHeader` and optionally the `AckedPacketHeader` from the underlying buffer.
```

src/packet/packet_structure.rs
```diff
@@ -1,28 +1,48 @@
...
-            delivery_method,
+            delivery,
+            ordering,
         }
     }
 
-    /// Unreliable. Packets can be dropped, duplicated or arrive without order.
+    /// Create a new unreliable packet by passing the receiver, data.
+    ///
+    /// Unreliable; Packets can be dropped, duplicated or arrive without order.
```
fhaynes:
```suggestion
    /// Unreliable: Packets can be dropped, duplicated or arrive without order.
```

src/packet/packet_structure.rs
```diff
@@ -32,43 +52,198 @@ impl Packet {
     ///
     /// Basically just bare UDP, free to be dropped, used for very unnecessary data, great for 'general' position updates.
```
fhaynes:
```suggestion
    /// Basically just bare UDP. The packet may or may not be delivered.
```

fhaynes: I changed this because I don't think we should suggest use cases in these comments, but instead in the docs.

src/packet/packet_structure.rs
```diff
@@ -32,43 +52,198 @@ impl Packet {
...
     ///
     /// *Details*
     ///
     /// |   Packet Drop   | Packet Duplication | Packet Order     | Packet Fragmentation | Packet Delivery |
     /// | :-------------: | :-------------:    | :-------------:  | :-------------:      | :-------------: |
     /// |       No        |      No            |      No          |      Yes             |       Yes       |
     ///
-    /// Basically this is almost TCP like without ordering of packets.
-    /// Receive every packet and immediately give to application, order does not matter.
+    /// Basically this is almost TCP-like without ordering of packets.
```
fhaynes:
```suggestion
    /// Basically this is almost TCP without ordering of packets.
```

src/packet/outgoing.rs
fhaynes: @TimonPost I think this can be an optimization in a second round if you want to make an issue for it. This PR has already been outstanding for a long time, and is ginormous; I'd prefer to get this merged as it is becoming difficult to code around it.

fhaynes:
> This PR is about 560 LoC issue related. Due to more added unit tests this PR is now over 50% unit 
> tests. The other changed code is about moving existing logic, which doesn't need to have a review.
> @jsfnl was fine with lots of tests and would more like to see 2000 lines PR with 1500 lines of unit 
> tests instead of 2000 lines PR with 0. You said that it was "ok" to have lots of tests in a PR.

> I did my best to keep it small, but if there was a limit of 200 lines it was impossible to add this 
> feature. 200 lines is really small. I can see that this works out in a more mature library. I think 550 
> LoC is acceptabel. If tests where also includes I would have left them away which is definitely not 
> plausible. They at least assure that those changes are right.

Yes, I understand. I was responding to the general case, not this PR specifically.

TimonPost:
bors r=fletcher

PR #155, opened 2019-03-13
Change CRC32 to CRC16 in `protocol_version`.

### LINKED ISSUE #152: Switch protocol version to CRC16.
Protocol version is currently a CRC32, we could save two bytes by making it a CRC16.

I saw [this](https://forums.intel.com/s/question/0D50P00003yyG94SAE/16bit-crc-vs-32bit-crc?language=en_US)

So basically a CRC32 can cover 500MBytes and a CRC16 can cover: 8kBytes.

- Protocol version is located in the [StandardHeader](https://github.com/amethyst/laminar/blob/master/src/packet/header/standard_header.rs#L14)
- CRC32 is calculated [here](https://github.com/amethyst/laminar/blob/master/src/protocol_version.rs)

daxpedda:
@jstnlef you pointed out in discord some other ideas for that issue.
Still open to other ideas that we can discuss in #152.

Fixes #152.

### DIFF
--- a/benches/packet_processing.rs
+++ b/benches/packet_processing.rs
@@ -57,7 +57,7 @@ fn process_packet_when_received(connection: &mut VirtualConnection, data: &[u8])
 /// visible externally
 fn standard_header_bytes(delivery_method: DeliveryMethod) -> Vec<u8> {
     let mut buffer = Vec::new();
-    buffer.write_u32::<BigEndian>(ProtocolVersion::get_crc32());
+    buffer.write_u16::<BigEndian>(ProtocolVersion::get_crc16());
     // Represents a standard `Packet`
     buffer.write_u8(0);
     buffer.write_u8(delivery_method as u8);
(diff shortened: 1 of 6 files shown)


TimonPost:
Looks good! Let's wait what the CD/CI says.

TimonPost:
Let's merge this!

TimonPost:
bors -r=TimonPost

daxpedda:
As far as I can see this is failing because of a test, that isn't part of the PR, not being formatted.

jstnlef:
Why did we try to merge this? Did we reach a decision? Did I miss something?

TimonPost:
It is at least an improvement to our current situation, I don't see the reason why it could not be merged.

jstnlef:
I mean I guess so. Though the expectation I have is that this should be changed immediately. We need to decide what we want sooner rather than later.

TimonPost:
After that this PR was created we came to a conclusion that we might want to take another approach on a protocol version and CRC16 implementation, I don't expect this to be ready this week. This PR does only a minor tweak and I don't see the reason not to merge it; it is an improvement on what we have right now.

PR #158, opened 2019-03-16
Fixed buffer bug

TimonPost:
I encountered a bug that I had encountered before. It is about the receive buffer. I can remember some discussion about that and we decided to change it back to a dynamic vector. However this won't work as described on [this page](https://stackoverflow.com/questions/26778657/rust-udpsocket-recv-from-fails-with-end-of-file-but-i-can-see-the-incoming-pa). 

I added a unit tests which was first failing but worked after the fix.

### DIFF
--- a/src/net/socket.rs
+++ b/src/net/socket.rs
@@ -159,30 +159,49 @@ impl Socket {
 
 #[cfg(test)]
 mod tests {
-    use crate::{Socket, Packet, Config};
+    use crate::{Config, Packet, Socket};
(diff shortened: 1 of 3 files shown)


fhaynes:
bors r=fhaynes

TimonPost:
bors r=fhaynes

jstnlef:
Looks good👍

TimonPost:
I am just going to merge this, the same fragment test issue we had before.

PR #160, opened 2019-03-18
Refactored some tests

TimonPost:
There where some problems with the current test I wanted to fix. 

- laminar-tester
The laminar tester had a lot of dirty code which could be formatted in a more elegant way, I did this by putting the command line arguments into a config type. 

- The integration tests
We had some problems with those, we could not see some statistics, packets were not asserted in a  correct way, it was a bit difficult to expand to other test scenario's, and for some reason, it failed the current CI. I introduced a `Client` and a `Server` which are abstracting away some logic. 

I added a type `ThroughputMonitor` which can monitor some throughput. Both the laminar-tester(binary) and the tests in `/test` are using this to monitor the throughput, that's why I added it behind a flag.

This PR is a step up to the docker tests.


src/bin/laminar-tester.rs
```diff
@@ -18,101 +19,123 @@ fn main() {
...
-                packet_throughput += 1;
+        match event_receiver.recv() {
+            Ok(SocketEvent::Packet(_)) => {
+                throughput.tick();
             }
-            Ok(_) => {}
             Err(e) => {
                 error!("Error receiving packet: {:?}", e);
             }
+            _ => {}
```
fhaynes: We should probably print something here just to know it happened

src/bin/laminar-tester.rs
```diff
@@ -18,101 +19,123 @@ fn main() {
...
         }
-        if second_counter.elapsed().as_secs() >= 10 {
-            second_counter = Instant::now();
-            debug!("Packet throughput is: {}", packet_throughput);
-            debug!("Total packets is: {}", packets_total_received);
-            packet_throughput = 0;
-        }
+
+        println!("{}", throughput);
```
fhaynes: Should we use the logging macros for this to keep it consistent?

src/bin/laminar-tester.rs
```diff
@@ -123,33 +146,32 @@ fn run_client(test_name: &str, destination: &str, endpoint: &str, pps: &str, tes
...
         sender
             .send(test_packet.clone())
             .expect("Unable to send a client packet");
+
         packets_sent += 1;
-        let now = Instant::now();
-        let d = now - start_time;
-        if d >= test_duration {
+
+        if start_time.elapsed() >= config.maximal_duration {
```
fhaynes: This should probably be an error, since we expect it to not exceed the max duration?

src/net/socket.rs
```diff
@@ -61,12 +62,12 @@ impl Socket {
...
                             Ok(_) => {}
                             Err(e) => {
-                                error!("Error sending SocketEvent: {:?}", e);
+                                error!("Error sending SocketEvent: {:?}", e.description());
                             }
                         };
                     }
                     None => {
-                        error!("Empty packet received");
+                        debug!("Empty packet received");
```
fhaynes: Do we expect to ever receive an empty packet in this test? If not, it should probably be an error.

src/net/socket.rs
```diff
@@ -77,7 +78,7 @@ impl Socket {
             // Now grab all the packets waiting to be sent and send them
             while let Ok(p) = self.packet_receiver.try_recv() {
                 if let Err(e) = self.send_to(p) {
-                    error!("There was an error sending packet: {:?}", e);
+                    println!("There was an error sending packet: {:?}", e);
```
fhaynes: This shouldn't be a println!, especially as it is not in the testing code. println! goes right to stdout, and doesn't respect the logging facilities. This is especially important for users that might send logs off the box and won't be watching the stdout/stderr of one process.

src/throughput.rs
```diff
@@ -0,0 +1,115 @@
...
+///
+/// Throughput is calculated at some duration.
+/// For each duration an entry is created to keep track of the history of throughput.
+///
+/// With this type you can calculate the average or get the total and last measured throughput.
+pub struct ThroughputMonitoring {
+    throughput_duration: Duration,
+    timer: Instant,
+    current_throughput: u32,
+    measured_throughput: Vec<ThroughputEntry>,
```
fhaynes: I found this name to be very confusing. Can we make it something like "throughput_measurements"?

src/throughput.rs
```diff
@@ -0,0 +1,115 @@
...
+    pub fn average(&self) -> u32 {
+        if self.measured_throughput.len() != 0 {
+            return self
+                .measured_throughput
+                .iter()
+                .map(|x| x.measured_throughput)
+                .sum::<u32>()
+                / self.measured_throughput.len() as u32;
+        }
+        0
```
fhaynes: Probably want to return a Result here, as this could obfuscate bugs/errors.

src/throughput.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    /// Reset the throughput history.
+    pub fn reset(&mut self) {
+        self.current_throughput = 0;
+        self.measured_throughput.clear();
+    }
+
+    /// Returns the last measured throughput.
+    pub fn last_throughput(&self) -> u32 {
+        self.measured_throughput.last().unwrap().measured_throughput
```
fhaynes: UNWRAP ALERT! WARNING! DANGER!

src/throughput.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /// Returns the totals measured throughput ticks.
+    pub fn total_measured(&self) -> u32 {
```
fhaynes: Can we name this "total_measured_ticks"?

src/throughput.rs
```diff
@@ -0,0 +1,115 @@
...
+            .map(|x| x.measured_throughput)
+            .sum::<u32>()
+    }
+}
+
+impl Debug for ThroughputMonitoring {
+    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        write!(
+            f,
+            "Current Throughput: {}, Elapsed Time: {:?}, Average Throughput: {}",
```
fhaynes: Since you are doing debug printing on the elapsed time, you may want to do `{:#?}` for pretty printing

src/throughput.rs
```diff
@@ -0,0 +1,115 @@
...
+            self.average()
+        )
+    }
+}
+
+impl Display for ThroughputMonitoring {
+    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        write!(
+            f,
+            "Current Throughput: {}, Elapsed Time: {:?}, Average Throughput: {}",
```
fhaynes: Same as above on the pretty printing

tests/common/client.rs
```diff
@@ -0,0 +1,72 @@
...
+use laminar::Socket;
+use std::net::SocketAddr;
+use std::thread;
+use std::thread::JoinHandle;
+use std::time::Duration;
+
+/// Represents a client to some endpoint.
+pub struct Client {
+    /// The sending timeout
+    pub timeout_sending: Duration,
```
fhaynes: This one might read better as "sending_timeout"

tests/common/server.rs
```diff
@@ -0,0 +1,145 @@
...
+        let serve_handle = thread::spawn(move || loop {
+            match packet_receiver.try_recv() {
+                Ok(result) => match result {
+                    SocketEvent::Packet(p) => {
+                        packet_assert(p);
+                        if throughput_monitor.tick() {
+                            if let Err(e) = events_tx.send(ServerEvent::Throughput(
+                                throughput_monitor.last_throughput(),
+                            )) {
+                                error!("Client can not send packet {:?}", e);
```
fhaynes: Should be `cannot`

tests/common/server.rs
```diff
@@ -0,0 +1,145 @@
...
+                            }
+                        }
+                    }
+                    _ => {
+                        if let Err(e) = events_tx.send(ServerEvent::SocketEvent(result)) {
+                            error!("Client can not send packet {:?}", e);
```
fhaynes: Should be `cannot`

src/bin/laminar-tester.rs
```diff
@@ -18,101 +19,123 @@ fn main() {
...
+impl From<clap::ArgMatches<'_>> for ClientConfiguration {
+    fn from(args: clap::ArgMatches<'_>) -> Self {
+        ClientConfiguration {
+            listen_host: args.value_of("LISTEN_ADDR").unwrap().parse().unwrap(),
+            destination: args.value_of("CONNECT_ADDR").unwrap().parse().unwrap(),
+            run_duration: Duration::from_secs(
+                args.value_of("SHUTDOWN_TIMER")
+                    .unwrap()
+                    .parse::<u64>()
+                    .unwrap(),
```
fhaynes: Unwrap alert

src/bin/laminar-tester.rs
```diff
@@ -18,101 +19,123 @@ fn main() {
...
 
-    let shutdown_timer = match st.parse::<u64>() {
-        Ok(parsed_st) => parsed_st,
-        Err(_) => {
-            error!("Invalid shutdown timer value");
-            exit(1);
+impl From<clap::ArgMatches<'_>> for ClientConfiguration {
+    fn from(args: clap::ArgMatches<'_>) -> Self {
+        ClientConfiguration {
+            listen_host: args.value_of("LISTEN_ADDR").unwrap().parse().unwrap(),
```
fhaynes: Unwrap alert

src/bin/laminar-tester.rs
```diff
@@ -18,101 +19,123 @@ fn main() {
...
+            destination: args.value_of("CONNECT_ADDR").unwrap().parse().unwrap(),
```
fhaynes: Unwrap alert

src/bin/laminar-tester.rs
```diff
@@ -18,101 +19,123 @@ fn main() {
...
+                args.value_of("SHUTDOWN_TIMER")
+                    .unwrap()
+                    .parse::<u64>()
+                    .unwrap(),
+            ),
+            packet_ps: args
+                .value_of("PACKETS_PER_SECOND")
+                .unwrap()
+                .parse()
+                .unwrap(),
```
fhaynes: Oo, double unwrap alert

src/bin/laminar-tester.rs
```diff
@@ -18,101 +19,123 @@ fn main() {
...
+            maximal_duration: Duration::from_secs(
+                args.value_of("TEST_DURATION").unwrap().parse().unwrap(),
```
fhaynes: Unwrap alert

src/bin/laminar-tester.rs
```diff
@@ -18,101 +19,123 @@ fn main() {
...
+            ),
+            test_name: String::from(args.value_of("TEST_TO_RUN").unwrap()),
```
fhaynes: And one more unwrap alert

src/bin/laminar-tester.rs
```diff
@@ -18,101 +19,123 @@ fn main() {
...
+#[derive(Clone)]
+struct ServerConfiguration {
+    listen_host: SocketAddr,
+    run_duration: Duration,
+}
+
+impl From<clap::ArgMatches<'_>> for ServerConfiguration {
+    fn from(args: clap::ArgMatches<'_>) -> Self {
+        ServerConfiguration {
+            listen_host: args.value_of("LISTEN_ADDR").unwrap().parse().unwrap(),
```
fhaynes: Unwrap...

src/bin/laminar-tester.rs
```diff
@@ -18,101 +19,123 @@ fn main() {
...
+            run_duration: Duration::from_secs(
+                args.value_of("SHUTDOWN_TIMER")
+                    .unwrap()
+                    .parse::<u64>()
+                    .unwrap(),
```
fhaynes: Man, mummies must be terrified of you.

src/bin/laminar-tester.rs
```diff
@@ -18,101 +19,123 @@ fn main() {
...
 
-fn process_server_subcommand(m: &clap::ArgMatches<'_>) {
-    let host = m.value_of("LISTEN_HOST").unwrap();
-    let port = m.value_of("LISTEN_PORT").unwrap();
-    let st = m.value_of("SHUTDOWN_TIMER").unwrap();
+struct ClientConfiguration {
+    listen_host: SocketAddr,
+    destination: SocketAddr,
+    run_duration: Duration,
+    packet_ps: u8,
```
jstnlef: Why just a byte? Seems the only place you're using it is as a `u64`. Might as well store it that way.

src/throughput.rs
```diff
@@ -0,0 +1,115 @@
...
+            return true;
+        } else {
+            self.current_throughput += 1;
+            return false;
+        }
+    }
+
+    /// Returns the average throughput over all throughput up-till now.
+    pub fn average(&self) -> u32 {
+        if self.measured_throughput.len() != 0 {
```
jstnlef: Instead of storing all of the intermediate measurements, it may make sense to calculate a running average rather than a total average. Would lessen the need to store all `ThroughputEntry`s.

src/bin/laminar-tester.rs
```diff
@@ -18,101 +19,123 @@ fn main() {
...
-        process_client_subcommand(m);
+        process_client_subcommand(m.to_owned());
     }
 }
 
-fn process_server_subcommand(m: &clap::ArgMatches<'_>) {
-    let host = m.value_of("LISTEN_HOST").unwrap();
-    let port = m.value_of("LISTEN_PORT").unwrap();
-    let st = m.value_of("SHUTDOWN_TIMER").unwrap();
+struct ClientConfiguration {
```
jstnlef: Like this change 👍

src/net/socket.rs
jstnlef: Seconded 👍

src/throughput.rs
jstnlef: Technically it should be `total_packets` or something to that effect.

src/net/socket.rs
TimonPost: We do and this should still be fixed, in case of fragmentation we could have None, however my other PR is removing a lot of this code and doesn't have to deal with this anymore.

src/throughput.rs
TimonPost: To me, it seems the same, I guess it better, let's go for it

TimonPost:
Why that many unwraps, 1) because they already existed 2) unwraps in tests are quite common, if something fails we need to stop the testing code; we could have expect though, I'll go for that

src/bin/laminar-tester.rs
TimonPost: Not sure what you mean with returning an error, the client should just stop sending when some time has elapsed.

src/throughput.rs
TimonPost: Does not have to be a result, if there are no measured throughputs the last measured throughput will be 0, because there are no measurements yet the average is also 0. Returning 0 instead of Result here is correct behavior.

src/throughput.rs
TimonPost: we can call it `total_packets` but what if we aren't monitoring packets? I like to have it a bit more generalized.

src/throughput.rs
TimonPost: Agree and disagree, agree because we aren't using the most out of this storage, disagree because we also store the time, of the measurement, this time will be useful to see throughput at certain times.

tests/unreliable_packets_test.rs
```diff
@@ -0,0 +1,133 @@
...
+        clients.push(client.run_instance(packet_factory, client_addr()));
+        info!("Client started.");
+    }
+
+    for client in clients {
+        client.wait_until_finished();
+        info!("Client finished.");
+    }
+
+    info!("Waiting 5 seconds");
```
TimonPost: 2 sec

src/bin/laminar-tester.rs
fhaynes: Ah, ok, so maximal duration is the amount of time the test should last, not a safety valve.

tests/unreliable_packets_test.rs
fhaynes: Sorry?

PR #161, opened 2019-03-20
Added in checking formatting and clippy

fhaynes:
Coverage will be done in another commit.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,22 @@
 pipeline {
     agent none
     stages {
+        stage('Check formatting and Clippy') {
+            environment {
+                CARGO_HOME = '/home/jenkins/.cargo'
+                RUSTUP_HOME = '/home/jenkins/.rustup'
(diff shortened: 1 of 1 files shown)


jstnlef:
Seems ok 👍

Jenkinsfile
```diff
@@ -1,6 +1,22 @@
...
+                CARGO_HOME = '/home/jenkins/.cargo'
+                RUSTUP_HOME = '/home/jenkins/.rustup'
+                RUSTFLAGS = "-D warnings"
+            }
+            agent {
+                label 'linux'
+            }
+            steps {
+                echo 'Checking formatting...'
+                sh '$CARGO_HOME/bin/cargo fmt -- --check'
```
LucioFranco: is it possible to break these out into different steps? I find it easier to parse the issues then.

fhaynes: Those are different steps. Each command is a step.

fhaynes: Oh, if you mean stage, sure.

PR #162, opened 2019-03-20
Adds in code coverage calculation via kcov.

fhaynes:
Adds in calculating code coverage via kcov when using Jenkins


jstnlef:
looks fine to me 👍

PR #165, opened 2019-03-26
Updated our docs

TimonPost:
I did some updates to our docs. If this PR is merged we can host our book somewhere, the question to @fhaynes is where? 

- I added diagrams. 
- I expanded the book with topics about reliability, arranging, and headers. 
- Removed some not written yet topics.

**This can be merged if #150 is merged**

One thing I like your toughts on also are use cases of the different reliability types described in `reliability.md`; I left a comment there with 'todo:  add use cases'.


LucioFranco:
The book should probably be hosted on gh-pages.

LucioFranco:
@TimonPost you want just the source in this PR, then we can have a CI job build it and deploy it.

docs/md_book/book/ayu-highlight.css
```diff
@@ -0,0 +1,71 @@
+/*
```
LucioFranco: we do not want this in the repo, this should be generated from CI and pushed to gh-pages

TimonPost:
Problem is, Jenkins will be triggered when a PR is made, if there are changes to the book, we want to re-build the book and publish the results to GH-pages.  However, if we push to GH-pages when a merge PR is created people could, in fact, remove the book and create a PR which will push an empty book to GH-pages and our book will be offline.

We want the book to be build once a merge to master is done, can we do this with bors or Jenkins?

PR #169, opened 2019-04-04
Added and more unit tests

TimonPost:
- Added more Unit Tests
- Renamed some unit tests
- Fixed some errors I discovered with unit tests.

### DIFF
--- a/src/net/virtual_connection.rs
+++ b/src/net/virtual_connection.rs
@@ -1,6 +1,6 @@
 use crate::{
     config::Config,
-    error::{ErrorKind, Result},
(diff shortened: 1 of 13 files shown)


src/net/virtual_connection.rs
```diff
@@ -635,7 +641,7 @@ mod tests {
     }
 
     #[test]
-    fn assure_right_processing_of_incoming_packets() {
+    fn assure_right_processing_of_incoming() {
```
fhaynes:
```suggestion
    fn assure_correct_processing_of_incoming() {
```

PR #170, opened 2019-04-11
Optional to have a config

TimonPost:


src/net/socket.rs
```diff
@@ -49,6 +51,13 @@ impl Socket {
         ))
     }
 
+    /// Configure the socket with the passed configuration.
+    pub fn with_config(mut self, config: Config) -> Socket {
```
jstnlef: I think I would prefer to have a `bind_with_config` method which is similar to the bind method today, as well as a `bind` method which calls it with `Config::default()`

examples/server_client.rs
```diff
@@ -8,8 +8,10 @@ use std::thread;
 
 const SERVER: &str = "127.0.0.1:12351";
 
+fn test() {}
```
LucioFranco: whats this for?

TimonPost: I taught it looked pretty!, no just kidding, it was from an other PR I was planning to make. But it should be removed.

examples/udp.rs
```diff
@@ -22,12 +22,8 @@ fn server_address() -> SocketAddr {
 
 /// This is an example of how to send data to an specific address.
 pub fn send_data() {
-    // you can change the config but if you want just go for the default.
-    let config = Config::default();
-
     // setup an udp socket and bind it to the client address.
```
fhaynes:
```suggestion
    // Setup a udp socket and bind it to the client address.
```

examples/simple_udp.rs
```diff
@@ -81,12 +81,8 @@ struct Server {
 impl Server {
     #[allow(unused_must_use)]
```
fhaynes: Can we remove this?

examples/server_client.rs
```diff
@@ -9,7 +9,7 @@ use std::thread;
 const SERVER: &str = "127.0.0.1:12351";
 
 fn server() -> Result<(), ErrorKind> {
-    let (mut socket, packet_sender, event_receiver) = Socket::bind(SERVER, Config::default())?;
+    let (mut socket, packet_sender, event_receiver) = Socket::bind(SERVER)?;
     let _thread = thread::spawn(move || socket.start_polling());
 
     println!("Listening for connections to {}", SERVER);
```
fhaynes: We should probably use a logging facility for this, `info!()` for example

examples/simple_udp.rs
```diff
@@ -149,12 +145,8 @@ struct Client {
 impl Client {
     #[allow(unused_must_use)]
```
fhaynes: Do we need this?

examples/server_client.rs
TimonPost: It is not related to my PR but I can throw it in.

fhaynes: Apologies, didn't even think about that. It's fine with me with either way. Resolving this.

PR #175, opened 2019-04-16
Fix reliable packets not realizing they were dropped

### LINKED ISSUE #174: [BUG] reliable_unordered is not reliable
See this example: https://gist.github.com/CosineP/52e5068375bd1ea11374fc4dac34c1d3

If packets are lost, they never make it to receiver (for example, i'm using `tc qdisc change dev eth0 root netem loss 10% 25%`)

this applies:
- in the wild (not a TC fluke)
- on linux and macos

CosineP:
Fixes #174

### DIFF
--- a/examples/server_client.rs
+++ b/examples/server_client.rs
@@ -8,8 +8,13 @@ use std::thread;
 
 const SERVER: &str = "127.0.0.1:12351";
 
+
+fn test() {
+
+}
+
 fn server() -> Result<(), ErrorKind> {
-    let (mut socket, packet_sender, event_receiver) = Socket::bind(SERVER, Config::default())?;
+    let (mut socket, packet_sender, event_receiver) = Socket::bind(SERVER)?;
     let _thread = thread::spawn(move || socket.start_polling());
 
     println!("Listening for connections to {}", SERVER);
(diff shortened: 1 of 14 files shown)


CosineP:
Hold up don't merge this yet.... even though it passes the unit test, it doesn't pass my integration test....

examples/server_client.rs
```diff
@@ -8,6 +8,8 @@ use std::thread;
 
 const SERVER: &str = "127.0.0.1:12351";
 
+fn test() {}
```
TimonPost: remove this, please!

examples/udp.rs
```diff
@@ -4,6 +4,7 @@
 //! 3. constructing the packet for sending.
 use laminar::{Config, Packet, Socket, SocketEvent};
 
+use crossbeam_channel::Sender;
```
TimonPost: can this file be restored to original

src/infrastructure/acknowlegement.rs
```diff
@@ -84,6 +84,26 @@ mod test {
         assert_eq!(handler.dropped_packets.len(), 0);
     }
 
+    #[test]
+    fn acking_500_packets_with_packet_drop() {
```
TimonPost: push the new test as we spoke about

TimonPost:
bors r=TimonPost

PR #176, opened 2019-04-17
Write + read headers properly for reliable packets

CosineP:
There were two problems:
    
    1. New packets were sent without the necessary headers including an
    updated index
    2. Acknowledgement was reading *its own* headers, instead of the foreign
    headers in calculating dropped vs acknowledged packets

This PR includes #175 because it hasn't been merged yet :/ let me know if i need to rebase it but it'd be a lot of work for nothing since we need to merge both

### DIFF
--- a/examples/udp.rs
+++ b/examples/udp.rs
@@ -4,7 +4,6 @@
 //! 3. constructing the packet for sending.
 use laminar::{Config, Packet, Socket, SocketEvent};
 
-use crossbeam_channel::Sender;
 use std::net::SocketAddr;
 use std::thread;
 
@@ -22,10 +21,15 @@ fn server_address() -> SocketAddr {
 }
 
 /// This is an example of how to send data to an specific address.
-pub fn send_data(sender: &Sender<Packet>) {
+pub fn send_data() {
+    // Setup a udp socket and bind it to the client address.
+    let (mut socket, packet_sender, _event_receiver) = Socket::bind(client_address()).unwrap();
+    let _thread = thread::spawn(move || socket.start_polling());
+
     let packet = construct_packet();
+
     // next send or packet to the endpoint we earlier putted into the packet.
-    sender.send(packet).unwrap();
+    packet_sender.send(packet).unwrap();
 }
 
 /// This is an example of how to receive data over udp on an specific socket address.
@@ -34,26 +38,25 @@ pub fn receive_data() {
     let (mut socket, _packet_sender, event_receiver) = Socket::bind(server_address()).unwrap();
     let _thread = thread::spawn(move || socket.start_polling());
 
-    let mut count = 0;
+    // Next start receiving.
+    let result = event_receiver.recv();
 
-    loop {
-        // Next start receiving.
-        let result = event_receiver.recv();
+    match result {
+        Ok(SocketEvent::Packet(packet)) => {
+            let endpoint: SocketAddr = packet.addr();
+            let received_data: &[u8] = packet.payload();
 
-        match result {
-            Ok(SocketEvent::Packet(packet)) => {
-                let endpoint: SocketAddr = packet.addr();
-                let received_data: &[u8] = packet.payload();
+            // you can here deserialize your bytes into the data you have passed it when sending.
 
-                // you can here deserialize your bytes into the data you have passed it when sending.
-                println!("Received {}", count);
-
-                count += 1;
-            }
-            Ok(_) => {}
-            Err(e) => {
-                println!("Something went wrong when receiving, error: {:?}", e);
-            }
+            println!(
+                "Received packet from: {:?} with length {}",
+                endpoint,
+                received_data.len()
+            );
+        }
+        Ok(_) => {}
+        Err(e) => {
+            println!("Something went wrong when receiving, error: {:?}", e);
         }
     }
 }
@@ -73,18 +76,4 @@ pub fn construct_packet() -> Packet {
 }
 
 // TODO: Use functions in example
-fn main() {
-    let handle = thread::spawn(move || {
-        receive_data();
-    });
-
-    let (mut socket, packet_sender, _event_receiver) = Socket::bind(client_address()).unwrap();
-    // Setup a udp socket and bind it to the client address.
-    let _thread = thread::spawn(move || socket.start_polling());
-
-    for i in 0..10000 {
-        send_data(&packet_sender);
-    }
-
-    handle.join();
-}
+fn main() {}
--- a/examples/udp.rs
+++ b/examples/udp.rs
@@ -4,9 +4,9 @@
 //! 3. constructing the packet for sending.
 use laminar::{Config, Packet, Socket, SocketEvent};
 
+use crossbeam_channel::Sender;
 use std::net::SocketAddr;
 use std::thread;
-use crossbeam_channel::Sender;
 
 /// The socket address of where the server is located.
 const SERVER_ADDR: &'static str = "127.0.0.1:12345";
@@ -46,12 +46,9 @@ pub fn receive_data() {
                 let received_data: &[u8] = packet.payload();
 
                 // you can here deserialize your bytes into the data you have passed it when sending.
-                println!(
-                    "Received {}",
-                    count
-                );
+                println!("Received {}", count);
 
-                count+=1;
+                count += 1;
             }
             Ok(_) => {}
             Err(e) => {
@@ -77,7 +74,6 @@ pub fn construct_packet() -> Packet {
 
 // TODO: Use functions in example
 fn main() {
-
     let handle = thread::spawn(move || {
         receive_data();
     });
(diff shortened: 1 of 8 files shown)


src/infrastructure/acknowlegement.rs
```diff
@@ -3,7 +3,7 @@ use crate::infrastructure::{ExternalAcks, LocalAckRecord};
 /// Type responsible for handling the acknowledgement of packets.
 pub struct AcknowledgementHandler {
     waiting_packets: LocalAckRecord,
-    their_acks: ExternalAcks,
+    our_acks: ExternalAcks,
```
jstnlef: Hrm. It looks like the ExternalAcks struct is conflating storing the remote sequence number with our own calculated version of the ack field. This implementation seems a bit confused.

CosineP:
what's amiss? The logic is definitely a little hard to follow since we're keeping track of our acknowledgement of foreign packets AND foreign acknowledgement of our packets, which is hard to name and keep in your head. i believe this was the cause of one of the bugs, which is why i renamed a variable to make it more clear imo

the other bug was a hard one to catch and could've happened any way

the logic pretty much matches gaffers article and reference implementation

src/infrastructure/acknowlegement.rs
CosineP: ExternalAcks keeps track of *our* receipt of the last 33 *foreign* packets, and that alone. if anything conflate its LocalAck between outgoing packets and foreign acknowledgement of our packets, but that's a fairly natural combo imo

src/net/socket.rs
```diff
@@ -131,13 +131,18 @@ impl Socket {
             }
 
             for payload in dropped_packets {
-                bytes_sent += self.send_packet(&packet.addr(), &payload)?;
+                bytes_sent += self.send_to(Packet::new(
```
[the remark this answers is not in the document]
CosineP: thank you. i will try in the morning

src/infrastructure/acknowlegement.rs
```diff
@@ -14,33 +14,33 @@ impl AcknowledgementHandler {
...
-            their_acks: Default::default(),
+            our_acks: Default::default(),
             dropped_packets: Vec::new(),
         }
     }
 }
 
 impl AcknowledgementHandler {
-    /// Returns the bit mask that contains the packets who are acknowledged.
+    /// Returns the bit mask that contains the packets who WE'VE receieved.
```
Hugal31: packets ~~who~~ WE'VE receieved ?

examples/udp.rs
```diff
@@ -4,6 +4,7 @@
 //! 3. constructing the packet for sending.
 use laminar::{Config, Packet, Socket, SocketEvent};
 
+use crossbeam_channel::Sender;
```
TimonPost: Reset file please, I think the rebase to the master will do this already.

src/infrastructure/acknowlegement.rs
TimonPost: I find the name `our_ackets` a bit confusing because now it seems like those acks are from our self but they are from 1 'them/their'

src/infrastructure/acknowlegement.rs
TimonPost:
```suggestion
    /// Returns the bit mask that contains the packets that we have received.
```

src/infrastructure/acknowlegement.rs
```diff
@@ -14,33 +14,33 @@ impl AcknowledgementHandler {
...
     pub fn bit_mask(&self) -> u32 {
-        self.their_acks.field
+        self.our_acks.field
     }
 
-    /// Returns the last acknowledged sequence number by the other endpoint.
+    /// Returns the last acknowledged sequence number WE'VE received.
```
TimonPost:
```suggestion
   /// Returns the last acknowledged sequence number that we received.
```

src/infrastructure/acknowlegement.rs
```diff
@@ -62,7 +62,8 @@ mod test {
         handler.seq_num = 40;
         handler.process_outgoing(vec![1, 2, 4].as_slice());
 
-        handler.process_incoming(40);
+        static ARBITRARY: u16 = 23;
```
TimonPost: does this add really that much of a difference?

src/infrastructure/acknowlegement.rs
```diff
@@ -73,24 +74,53 @@ mod test {
...
+
+        let mut drop_count = 0;
+
+        for i in 0..2000 {
+            handler.process_outgoing(vec![1, 2, 3].as_slice());
+            handler.seq_num = i;
+
+            // dropping every 4th with modulo's
+            if i % 4 == 0 {
+                println!("Dropping packet: {}", drop_count);
```
TimonPost: change to `debug!`

src/infrastructure/external_ack.rs
```diff
@@ -180,14 +184,6 @@ mod test {
         acks.ack(6);
         acks.ack(4);
         assert_eq!(acks.last_seq, 6);
-        assert_eq!(
```
TimonPost: This might be removed, I have one concern to this which is that people will never know what `0b110010` means. The previous code was more cleaner.

src/infrastructure/acknowlegement.rs
CosineP: The acks *are* from ourself! They are our acknowledgements of foreign packet receipt

src/infrastructure/acknowlegement.rs
CosineP: I had to add two arguments, and it made sense to me this way

src/net/socket.rs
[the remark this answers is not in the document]
CosineP: can you help me write this test? im not sure how...

src/infrastructure/external_ack.rs
CosineP: the bit shifts, really? what if i commented it, it seems clearer to me this way. but if you prefer i will reset it

jstnlef:
Sorry I didn't get back to you sooner. It's been a crazy couple of weeks for me both at work and outside of work. Let me elaborate on what I was trying to get at the other day.
First, when I say "confused", what I mean is that I think somewhere in the refactor to get the streams logic in, we lost the actual acking algorithm we are trying convey. 

For example, we are currently not calculating the ack_bits on send. We are calculating them on receive. I believe this to be a mistake. Really, what should happen is that right before we're about to send a packet to the other side is when we want to calculate the ack_bits. This ensures that the values we put in the outgoing packet's header will be the most up to date they can be at time of send. And to your point @CosineP, you're absolutely right that the ack_bits are generated based off of our receiving of remote packets.

Another thing I worry about is our algorithm around generating the ack_bits themselves. For example, why the seemingly arbitrary `32000` value as well as multiple if conditions? The Gaffer C implementation uses this algorithm to generate the ack_bits:
```
void reliable_sequence_buffer_generate_ack_bits( struct reliable_sequence_buffer_t * sequence_buffer, uint16_t * ack, uint32_t * ack_bits )
{
    reliable_assert( sequence_buffer );
    reliable_assert( ack );
    reliable_assert( ack_bits );
    *ack = sequence_buffer->sequence - 1;
    *ack_bits = 0;
    uint32_t mask = 1;
    int i;
    for ( i = 0; i < 32; ++i )
    {
        uint16_t sequence = *ack - ((uint16_t)i);
        if ( reliable_sequence_buffer_exists( sequence_buffer, sequence ) )
            *ack_bits |= mask;
        mask <<= 1;
    }
}

int reliable_sequence_buffer_exists( struct reliable_sequence_buffer_t * sequence_buffer, uint16_t sequence )
{
    reliable_assert( sequence_buffer );
    return sequence_buffer->entry_sequence[ sequence % sequence_buffer->num_entries ] == (uint32_t) sequence;
}
```
which seems pretty straightforward to me. Anyway, the next couple of days (most likely Saturday) I'm going to get out my scalpel and rearrange some stuff to simplify this and have it follow the Gaffer implementation more closely and see if that clears up some stuff.

TimonPost:
Awesome, I'd like to see this acking to get better, it's the last piece to make laminar awesome and solid. I said this a vew times, but the implementation we currently have is from [here](https://github.com/acmcarther/gaffer_udp)

src/infrastructure/external_ack.rs
TimonPost: I am a bit unfamiliar with that too, however, a random number might be more unclear, What does it mean, at least if you get the general idea of bit-shifting it will be easier to understand.

Hugal31: What about "0b01 | 0b10" ?

src/infrastructure/acknowlegement.rs
TimonPost: Thanks for clarifying your vision, let's leave it then.

CosineP:
i opted to create a new struct for a waiting packet instead of reusing Packet because the `delivery` and `addr` packets are redundant, but it's a little bit of duplication so i can change it to `Packet` if you want

TimonPost:
Thanks for your PR, great work!

TimonPost:
bors r=TimonPost, jstnlef

PR #177, opened 2019-04-22
Correct/Simplify the Ack algorithm

jstnlef:
Honestly, I have a lot more tests to write before this goes out the door but I wanted to get some initial feedback on what I have currently.

Key highlights are -
1. I rewrote the sequence buffer to handle some cases we were explicitly not handling before. For example, if a sequence number is too old, we need to ensure that we don't place it into the buffer. Also, the perfectly valid sequence number `u16::max` was being used as a "null" value which means that we would have incorrect behavior.
2. I redid the ack-ing code to be a bit simpler and easy to follow.
3. I am now no longer counting packets as dropped until we haven't received an ack for those packets in 32+ sequence numbers.

I also added some descriptive comments around why I'm calculating the fields as I am.


src/infrastructure/acknowledgement.rs
```diff
@@ -0,0 +1,241 @@
+use crate::packet::OrderingGuarantee;
+use crate::sequence_buffer::SequenceBuffer;
+use crate::packet::SequenceNumber;
+use std::collections::HashMap;
+
+const REDUNDANT_PACKET_ACKS_SIZE: u16 = 32;
+
+/// Responsible for handling the acknowledgement of packets.
```
TimonPost:
```suggestion
/// Responsible for handling the acknowledgment of packets.
```

src/infrastructure/acknowledgement.rs
```diff
@@ -0,0 +1,241 @@
...
+    // Local sequence number which we'll bump each time we send a new packet over the network
+    sequence_number: SequenceNumber,
+    // Using a Hashmap to track every packet we send out so we can ensure that we can resend when
+    // dropped.
+    sent_packets: HashMap<u16, SentPacket>,
+    // However, we can only reasonably ack up to REDUNDANT_PACKET_ACKS_SIZE + 1 packets on each
+    // message we send so this should be REDUNDANT_PACKET_ACKS_SIZE large.
+    received_packets: SequenceBuffer<ReceivedPacket>,
+
+    pub dropped_packets: Vec<SentPacket>,
```
TimonPost: Since this is public we should have a comment here

TimonPost: Maybe it would be a good idea to have a function `drain_dropped_packets()` which calls `drain` on the vector, just to have better encapsulation. Currently, you're calling this in `virtual_connection.rs` line 373 where you do the drain there.

src/infrastructure/acknowledgement.rs
jstnlef: Heh surprised to see this here since it's common to include the e outside of the US. I can change it throughout if you want.

src/infrastructure/acknowledgement.rs
jstnlef: Love that idea 👍

src/infrastructure/acknowledgement.rs
TimonPost: hrrmm, I didn't even know that there were multiple ways to write this, my grammar checkers just came up with this. Just do what you think that's correct English :)

jstnlef: I'll change it across the board with the new spelling. 👍

src/infrastructure/acknowledgment.rs
```diff
@@ -0,0 +1,249 @@
...
+
+    /// Returns the last sequence number received from the remote host (+1)
+    pub fn remote_sequence_num(&self) -> SequenceNumber {
+        self.received_packets.sequence_num()
+    }
+
+    /// Returns the ack_bitfield corresponding to which of the past 32 packets we've
+    /// successfully received.
+    pub fn ack_bitfield(&self) -> u32 {
+        let most_recent_remote_seq_num: u16 = self.remote_sequence_num().wrapping_sub(1);
```
TimonPost: `last_remote_seq_num`?

src/infrastructure/acknowledgment.rs
```diff
@@ -0,0 +1,249 @@
...
+                // We send them a packet
+                other.process_incoming(i, handler.remote_sequence_num(), handler.ack_bitfield());
+                // Skipped: other.process_outgoing
+                // And it makes it back
+                handler.process_incoming(i, other.remote_sequence_num(), other.ack_bitfield());
+            }
+        }
+
+        // TODO: Is this what we want?
+        //        assert_eq!(handler.dropped_packets.len(), 25);
```
TimonPost: I think this is exactly what we want if we have 100 packets and we divide that by 4 we will get 25. So we should have 25 dropped packets over 100 packets. If you don't trust the modulo operation please try the following code out:

```rust

let mut counter = 1;
        for i in 0..100 {
            handler.process_outgoing(vec![1, 2, 3].as_slice(), OrderingGuarantee::None);
            handler.sequence_number = i;

            // dropping every 4th with modulo's
            if counter == 4 {
                println!("Dropping packet: {}", drop_count);
                drop_count += 1;

                counter = 1;
            } else {
                // We send them a packet
                other.process_incoming(i, handler.remote_sequence_num(), handler.ack_bitfield());
                // Skipped: other.process_outgoing
                // And it makes it back
                handler.process_incoming(i, other.remote_sequence_num(), other.ack_bitfield());

                counter += 1;
            }
        }
```
This will drop every fourth packet for sure.

TimonPost:
The original acking code has some nice tests by the way which you could use as a reference for this implementation.

src/infrastructure/acknowledgment.rs
jstnlef: I wanted to make this more specific in that it should be the sequence number that is furthest ahead rather than the last one we've received.

src/infrastructure/acknowledgment.rs
jstnlef: This goes back to the discussion we were having the other day on discord (plus I know how modulo works :P). The reason why we would only get 17 dropped packets in this case is that the algorithm waits until it sees 32 packets that haven't acked that sequence number. So at received packet 100, which is the last packet received from the remote host, we have an ack which looks like `01110111011101110111011101110111`. <- This is where those extra 8 packets are. We haven't yet determined that they have been dropped because they haven't "fallen off the end"

examples/simple_udp.rs
```diff
@@ -160,7 +160,8 @@ impl Client {
         match serialized {
             Ok(raw_data) => {
                 self.packet_sender
-                    .send(Packet::reliable_unordered(server_address(), raw_data));
+                    .send(Packet::reliable_unordered(server_address(), raw_data))
+                    .expect("Should be fine");
```
fhaynes: What does this mean, exactly?

src/net/socket.rs
```diff
@@ -187,6 +178,15 @@ impl Socket {
         let bytes_sent = self.socket.send_to(payload, addr)?;
         Ok(bytes_sent)
     }
+
+    // In the presence of a link conditioner, we may not want to packet each time.
```
fhaynes: to packet each time?

jstnlef:
I feel pretty good about this at this point. Would like to see what you guys think.

jstnlef:
Not a bad idea. I'll see what I can do tonight 👍

jstnlef:
Hrm. I think it would take a bit to get this set up as a lazy iterator. We can take it as a future improvement I think 👍

TimonPost:
Right, let's get this merged and create an issue for it.

jstnlef:
Ticket here: https://github.com/amethyst/laminar/issues/182

PR #178, opened 2019-04-24
Removed unneeded error logging

fhaynes:


src/net/socket.rs
```diff
@@ -67,7 +67,7 @@ impl Socket {
         loop {
             // First we pull any newly arrived packets and handle them
             if let Err(e) = self.recv_from() {
-                error!("Error receiving packet: {:?}", e);
+
```
TimonPost: we have an empty block here?

fhaynes: Fixed

PR #179, opened 2019-04-24
Timeout Event

TimonPost:
- Fixed the sending of the TimeoutEvent
- Added tests to prevent this in the future

### DIFF
--- a/examples/server_client.rs
+++ b/examples/server_client.rs
@@ -8,8 +8,13 @@ use std::thread;
 
 const SERVER: &str = "127.0.0.1:12351";
 
+
+fn test() {
+
+}
+
 fn server() -> Result<(), ErrorKind> {
-    let (mut socket, packet_sender, event_receiver) = Socket::bind(SERVER, Config::default())?;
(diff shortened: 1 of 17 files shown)


examples/server_client.rs
```diff
@@ -8,6 +8,8 @@ use std::thread;
 
 const SERVER: &str = "127.0.0.1:12351";
 
+fn test() {}
```
jstnlef: this thing keeps coming back :P

src/net/socket.rs
```diff
@@ -78,20 +78,25 @@ impl Socket {
             }
 
             // Finally check for idle clients
-            self.handle_idle_clients();
+            if let Err(e) = self.handle_idle_clients() {
+                error!("Encountered an error when sending TimeoutEvent {:?}", e);
```
jstnlef: Seems fine. May want to add a `:` to make it consistent with the rest of them.

examples/server_client.rs
TimonPost: Sadly yes :(

PR #181, opened 2019-04-27
Remove accidental infinite recursion in Socket::send_to

jstnlef:
This was discovered by @zesterer on the Discord channel. Basically, while the issue exists on master, it was exercised by the new acking code in this PR https://github.com/amethyst/laminar/pull/177. The new acking code, when it doesn't receive a message back from the remote host, has no way of knowing that any packets got through. So in that case, we will resend messages. I'm not entirely sure why the original acking code didn't do this, however.

Instead, what I've put together here will call `process_outgoing` on the packet requested to be sent and all of the dropped packets and concatenate them all into a single Vec. From that point on, we treat them all as separate new messages needed to be sent.

### DIFF
--- a/src/net/socket.rs
+++ b/src/net/socket.rs
@@ -407,6 +407,8 @@ mod tests {
                 _ => None,
             })
             .collect();
-        assert_eq!(sent_events, vec![0, 1, 35]);
+        assert_eq!(sent_events.len(), 3);
+        // The order will be guaranteed in a future PR.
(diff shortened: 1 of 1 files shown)


jstnlef:
Added the test. One interesting thing I found is that because we're using a hashmap to store the waiting packets, we don't have a predictable send order. The new acking PR ensures that, at the very least, we are sending packets predictably.

PR #191, opened 2019-06-13
socket: Allow manual polling

BourgondAries:
The problem with the current polling loop is that it does not wait,
causing CPU usage to skyrocket for no good reason. In a game we would
like to run `poll` during every game tick, as there is no point in
running any faster than that.

### DIFF
--- a/src/net/socket.rs
+++ b/src/net/socket.rs
@@ -22,6 +22,11 @@ pub struct Socket {
     packet_receiver: Receiver<Packet>,
 }
 
+enum UdpSocketState {
+    Empty,
+    MaybeMore,
+}
+
 impl Socket {
     /// Binds to the socket and then sets up `ActiveConnections` to manage the "connections".
     /// Because UDP connections are not persistent, we can only infer the status of the remote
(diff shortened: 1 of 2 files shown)


BourgondAries:
Looks like the test failure is caused by `Caused: org.kohsuke.github.HttpException: Server returned HTTP response code: -1, message: 'null' for URL: https://api.github.com/repos/amethyst/laminar`

Should be unrelated to the PR

BourgondAries:
>You say that a server tick should manually poll for events, 

That is my use-case, I think it's up to the user of the library to decide how to poll, or to wait, or spawn a thread to poll.

>however, I do have a question on that. Like if you only receive once a tick, all the messages that will be sent between the ticks will not be received. That is how UDP works. So how do you plan to fix that with a single poll each tick?

That is not how UDP works as far as I know. The socket is still listening and the kernel buffers the messages. See `sysctl net.core.rmem_default`.
The test `manual_polling_socket` confirms this.

src/net/socket.rs
```diff
@@ -65,26 +70,36 @@ impl Socket {
...
-                        ErrorKind::IOError(ref e) if e.kind() == io::ErrorKind::WouldBlock => {}
-                        _ => error!("There was an error sending packet: {:?}", e),
-                    }
-                }
+            self.manual_poll()?;
+        }
+    }
+
+    /// Process any inbound/outbound packets and handle idle clients
+    pub fn manual_poll(&mut self) -> Result<()> {
```
jstnlef: The way this is currently set up, you'll never return an Err result so you should probably remove the result type.

src/net/socket.rs
```diff
@@ -65,26 +70,36 @@ impl Socket {
...
-
-            // Now grab all the packets waiting to be sent and send them
-            while let Ok(p) = self.packet_receiver.try_recv() {
-                if let Err(e) = self.send_to(p) {
-                    match e {
-                        ErrorKind::IOError(ref e) if e.kind() == io::ErrorKind::WouldBlock => {}
-                        _ => error!("There was an error sending packet: {:?}", e),
-                    }
-                }
+            self.manual_poll()?;
```
jstnlef: This mustn't return an Err result or else the polling loop is broken

BourgondAries: This breaks the API, the previous implementation did also return a result, was the previous wrong?
Anyhow, I've removed it and changes the examples.

BourgondAries:
Should be good now.

jstnlef:
This looks good now 👍 . I'll fix our jenkins situation when I get home tonight.

TimonPost:
Seems fine let's merge this one

PR #192, opened 2019-06-13
Move time access to public boundaries

BourgondAries:
NOTE: This series depends on the manual polling patch series, I'm not sure how else I could split these up. See https://github.com/amethyst/laminar/pull/191

Accessing `Instant::now()` from within makes this library difficult to
test, especially when testing another system together with this library
you want to be able to precisely control time.

This also prevents tests from being flaky (the test using sleep has been replaced with an exact nanosecond-granularity check).

Using the provided looping poll from socket will still use its own time
so it's backward-compatible.

Some other parts such as "throughput" still use Instant but that should
probably be swapped out for a provided time as well.

### DIFF
--- a/src/infrastructure/congestion.rs
+++ b/src/infrastructure/congestion.rs
@@ -35,9 +35,9 @@ impl CongestionHandler {
     ///
     /// This will insert an entry which is used for keeping track of the sending time.
     /// Once we process incoming sequence numbers we can calculate the `RTT` time.
-    pub fn process_outgoing(&mut self, seq: u16) {
+    pub fn process_outgoing(&mut self, seq: u16, time: Instant) {
         self.congestion_data
-            .insert(seq, CongestionData::new(seq, Instant::now()));
+            .insert(seq, CongestionData::new(seq, time));
     }
 }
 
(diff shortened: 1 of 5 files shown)


BourgondAries:
I agree with the argument position. Will do.

BourgondAries:
@jstnlef Ok, should be good now.

jstnlef:
Looks like the windows build is broken
```
failures:


---- net::connection::tests::connection_timed_out stdout ----

thread 'net::connection::tests::connection_timed_out' panicked at 'assertion failed: `(left == right)`

  left: `0`,

 right: `10`', src\net\connection.rs:101:9

note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

```

BourgondAries:
@jstnlef Unfortunately I don't use windows so I can't test this. I don't see how this can be broken on windows unless the timer resolution is less there.

jstnlef:
That very well could be the issue... (sigh) For the fun of it, could you try to add a `Duration::from_nanos(1)` to the `now + wait`. I'm wondering if it's a boundary issue on Windows for that API.

BourgondAries:
Adding a nanos doesn't work. Next up, a micro

jstnlef:
Jenkins is being such a pain. I'm alright with this now 👍

PR #194, opened 2019-06-14
DoS attack mitigation

BourgondAries:
This fixes issue 187. See last commit in the series for more detail.

### DIFF
--- a/src/net/socket.rs
+++ b/src/net/socket.rs
@@ -1,3 +1,4 @@
+use crate::either::Either::{Left, Right};
 use crate::{
     config::Config,
     error::{ErrorKind, Result},
@@ -204,9 +205,16 @@ impl Socket {
 
                 let connection =
                     self.connections
-                        .get_or_insert_connection(address, &self.config, time);
+                        .get_or_create_connection(address, &self.config, time);
 
-                connection.process_incoming(received_payload, &self.event_sender, time)?;
+                match connection {
+                    Left(existing) => {
+                        existing.process_incoming(received_payload, &self.event_sender, time)?;
+                    }
+                    Right(mut anonymous) => {
+                        anonymous.process_incoming(received_payload, &self.event_sender, time)?;
+                    }
+                }
             }
             Err(e) => {
                 if e.kind() != io::ErrorKind::WouldBlock {
@@ -235,6 +243,32 @@ impl Socket {
             true
         }
     }
+
+    #[cfg(test)]
+    pub fn connection_count(&self) -> usize {
+        self.connections.count()
+    }
+
+    #[cfg(test)]
+    pub fn forget_all_incoming_packets(&mut self) {
+        loop {
+            match self.socket.recv_from(&mut self.recv_buffer) {
+                Ok((recv_len, _address)) => {
+                    if recv_len == 0 {
+                        panic!("Received data too short");
+                    }
+                    let received_payload = &self.recv_buffer[..recv_len];
+                }
+                Err(e) => {
+                    if e.kind() != io::ErrorKind::WouldBlock {
+                        panic!("Encountered an error receiving data: {:?}", e);
+                    } else {
+                        return;
+                    }
+                }
+            }
+        }
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -26,6 +26,7 @@ travis-ci = { repository = "amethyst/laminar", branch = "master" }
 byteorder = "1.2"
 crc = "1.8"
 crossbeam-channel = "0.3"
+either = "1.5.2"
```
jstnlef: Must we add this? I really dislike adding small dependencies that can be written here in less than 15 mins

src/net/socket.rs
```diff
@@ -103,17 +120,18 @@ impl Socket {
...
 
     // Serializes and sends a `Packet` on the socket. On success, returns the number of bytes written.
-    fn send_to(&mut self, packet: Packet) -> Result<usize> {
-        let connection = self
-            .connections
-            .get_or_insert_connection(packet.addr(), &self.config);
+    fn send_to(&mut self, time: Instant, packet: Packet) -> Result<usize> {
+        let connection =
+            self.connections
+                .get_or_insert_connection(packet.addr(), time, &self.config);
```
jstnlef: If we do it this way, we will be actively losing state of the first few messages a remote client has sent us because we're creating a new VirtualConnection here. Put in a different way, we won't be able to ack the first few messages which could cause problems.

src/net/socket.rs
```diff
@@ -162,20 +181,29 @@ impl Socket {
...
+                    self.connections
+                        .get_or_create_connection(address, time, &self.config);
 
-                connection.process_incoming(received_payload, &self.event_sender)?;
+                match connection {
+                    Left(existing) => {
+                        existing.process_incoming(received_payload, time, &self.event_sender)?;
+                    }
+                    Right(mut anonymous) => {
+                        anonymous.process_incoming(received_payload, time, &self.event_sender)?;
```
jstnlef: This goes back to my earlier comment. Any processing of `anonymous` will be promptly throw out because we don't store it anywhere.

Cargo.toml
BourgondAries: I honestly think Either should be part of std, I don't like small dependencies either but everyone's familiar with Either, hence why. Can easily be replaced by a little enum though.

jstnlef: Please do :)

src/net/socket.rs
BourgondAries: Should we ack the first message? Why not let the client retry sending a few times instead? If we add the connection to the connection table we get the DoS problem again.

src/net/socket.rs
BourgondAries: The only side-effect this code has is that it puts a packet on the event queue. Isn't that ok? We do want to drop the anonymous connection because we don't want to be attacked. I'm not sure how we can do acking and other stream-like operations without adding this connection to some table.

jstnlef: Maybe worth tabling this, for now, in order to get a proper handshake implementation. What do you think?

jstnlef: Alternatively, can you add a test that proves that having the client resend a few times at the beginning doesn't cause any significant issues?

BourgondAries: I've added a test that shows a failing client that eventually gets added.

jstnlef: Sure but what about the state of that connection? Can we prove that we aren't missing messages we're expected to ack

jstnlef: You're current test doesn't exercise the reliable flow

BourgondAries: What is it that you'd like to see? Any reliable messages before the server has performed `send_to` will not be stored or acked. The client will resend them if within the timeout window. Should I check that the client resends these?

BourgondAries: We will be missing those messages, but since the client does not receive an ack, it will resend.

jstnlef: So, basically the behavior you're suggesting is, in the reliable only case, these things will be re-sent so you are guaranteed duplicates. I'm not positive we want this behavior. I'll have to think more on this.

BourgondAries: Don't we simply not put packets on the queue we've seen before (by using the sequence number)?

jstnlef:
Failing with a clippy error.

src/net/socket.rs
jstnlef: Not for unreliable currently.

BourgondAries: I mean for reliable packets. If the receiver compares the sequence number to see if it has already acked, then it can ack again but not put the packet on the queue, thus not getting any duplicates from the user's point of view.

jstnlef: The thing is, when those packets are re-sent, they're re-sent with a new sequence number so the user will necessarily get these packets twice

BourgondAries: Why are they sent with a new sequence number? If we store them in the not-yet-acked packages list together with their sequence number we can reuse the sequence number.

jstnlef: This has to do with how our ack-ing code works. I urge you to take a look at it in `acknowledgement.rs`. The short of it, though, is that with each packet we send the most recently received sequence number and a 32 bit mask representing the acknowledgement status of the previous 32 packets prior. If we were to reuse sequence numbers, this scheme wouldn't work.

BourgondAries: So should we just resend and accept duplicates? Ideally, the receiver should be idempotent with regards to the packets it received.

BourgondAries: The new test shows that clients not in the server's connection table will make the client simply not send the previously dropped packets.

BourgondAries: Scratch that, forgot a manual_poll, seems like we're not properly re-sending.

BourgondAries:
Should be good now, but we do need to send heartbeats to flush the stream and make everything tick.

BourgondAries:
I've removed the ordered and sequences tests since those are part of https://github.com/amethyst/laminar/pull/199, when this is merged, those tests should just work.

PR #196, opened 2019-06-15
Socket::get_port()

Ploppz:
Tiny PR. Not sure about the `.expect()` - is it ok? I'm not sure in what situations it would fail.

### DIFF
--- a/src/net/socket.rs
+++ b/src/net/socket.rs
@@ -166,6 +166,10 @@ impl Socket {
     pub fn set_link_conditioner(&mut self, link_conditioner: Option<LinkConditioner>) {
         self.link_conditioner = link_conditioner;
     }
+    /// Get port of socket
+    pub fn local_addr(&self) -> Result<SocketAddr> {
+        Ok(self.socket.local_addr()?)
+    }
 
     /// Iterate through all of the idle connections based on `idle_connection_timeout` config and
     /// remove them from the active connections. For each connection removed, we will send a
@@ -866,5 +870,13 @@ mod tests {
 
         // 101 because we have 0..100 and 255 from the dummies
         assert_eq![101, send_many_packets(Some(255))];
+    }
+
+    #[test]
+    fn local_addr() {
+        let port = 40000;
+        let socket =
+            Socket::bind(format!("127.0.0.1:{}", port).parse::<SocketAddr>().unwrap()).unwrap();
+        assert_eq!(port, socket.local_addr().unwrap().port());
     }
 }


BourgondAries:
Please implement a test for this.

BourgondAries:
I think the expect is fine, here's the relevant code from std:
```
fn sockname<F>(f: F) -> io::Result<SocketAddr>
    where F: FnOnce(*mut c::sockaddr, *mut c::socklen_t) -> c_int
{
    unsafe {
        let mut storage: c::sockaddr_storage = mem::zeroed();
        let mut len = mem::size_of_val(&storage) as c::socklen_t;
        cvt(f(&mut storage as *mut _ as *mut _, &mut len))?;
        sockaddr_to_addr(&storage, len as usize)
    }
}

pub fn sockaddr_to_addr(storage: &c::sockaddr_storage,
                    len: usize) -> io::Result<SocketAddr> {
    match storage.ss_family as c_int {
        c::AF_INET => {
            assert!(len as usize >= mem::size_of::<c::sockaddr_in>());
            Ok(SocketAddr::V4(FromInner::from_inner(unsafe {
                *(storage as *const _ as *const c::sockaddr_in)
            })))
        }
        c::AF_INET6 => {
            assert!(len as usize >= mem::size_of::<c::sockaddr_in6>());
            Ok(SocketAddr::V6(FromInner::from_inner(unsafe {
                *(storage as *const _ as *const c::sockaddr_in6)
            })))
        }
        _ => {
            Err(Error::new(ErrorKind::InvalidInput, "invalid argument"))
        }
    }
}
```

Considering a UDP socket is only AF_INET or AF_INET6, this should be safe. Or the OS fails on getsockname, which is unlikely.

jstnlef:
Pinging this

Ploppz:
Sorry, I have been meaning to get back to this and I agree with your suggestion. I'm just quite busy now for a few more days. I will change it when I have some time.

jstnlef:
No worries 👍

Ploppz:
Tiny change but is this something like you imagined? Or would it be better to store the `SocketAddr` at creation (`bind`)? I mean, I'm unsure when `local_addr()` can fail. Seeing as we create `UdpSocket` with`addresses: A` where `A: ToSocketAddr`.

jstnlef:
`self.socket.local_addr()` should do what you want :)

Ploppz:
ah of course >_> so I'll just make that public

jstnlef:
erm. I'd still prefer the method. That way we don't expose any of the internals of the struct.
```
pub fn local_addr(&self) -> Result<SocketAddr> {
    self.socket.local_addr()
}
```

jstnlef:
Just needs a rebase now :)

BourgondAries:
Nit: Add a newline before `  #[test]`

Ploppz:
Almost forgot to say: the rebase is done.

PR #199, opened 2019-06-16
[bugfix] Correctly store arranging numbers for ordered and sequenced packets

BourgondAries:
Previously, ordered packets would simply attain a new item identifier
from the stream generator if they were dropped. This patch ensures that
the not-yet-acked packets are stored with this value so that they use
that same value when they are re-sent.

### DIFF
--- a/src/net/socket.rs
+++ b/src/net/socket.rs
@@ -247,12 +247,12 @@ impl Socket {
     }
 
     #[cfg(test)]
-    pub fn connection_count(&self) -> usize {
+    fn connection_count(&self) -> usize {
         self.connections.count()
     }
 
     #[cfg(test)]
-    pub fn forget_all_incoming_packets(&mut self) {
+    fn forget_all_incoming_packets(&mut self) {
         loop {
             match self.socket.recv_from(&mut self.recv_buffer) {
                 Ok((recv_len, _address)) => {
--- a/src/net/socket.rs
+++ b/src/net/socket.rs
@@ -157,6 +157,7 @@ impl Socket {
                     DeliveryGuarantee::Reliable,
                     // This is stored with the dropped packet because they could be mixed
                     waiting_packet.ordering_guarantee,
+                    waiting_packet.item_identifier,
                     time,
                 )
             })
@@ -166,6 +167,7 @@ impl Socket {
             packet.payload(),
             packet.delivery_guarantee(),
             packet.order_guarantee(),
+            None,
             time,
         )?;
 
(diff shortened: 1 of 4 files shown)


src/packet.rs
```diff
@@ -12,6 +12,18 @@ pub use self::outgoing::{Outgoing, OutgoingPacket, OutgoingPacketBuilder};
 pub use self::packet_reader::PacketReader;
 pub use self::packet_structure::Packet;
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum ItemIdentifier {
```
jstnlef: Erm.. Couldn't this be an `Option<SequenceNumber>` instead of a new type?

BourgondAries: It makes it hard to see at call-sites what you're sending in, especially if you have two
```
call(
    ...,
    Some(1),
    Some(1),
)
```

src/net/socket.rs
```diff
@@ -212,6 +214,32 @@ impl Socket {
             true
         }
     }
+
+    #[cfg(test)]
+    pub fn connection_count(&self) -> usize {
```
TimonPost: I prefer keeping tests inside the test module and test the API from there.

BourgondAries: It's not a test, just a function used only in tests.

TimonPost: so why not move that function to the test module? You can call into private modules there as well.

src/net/socket.rs
```diff
@@ -212,6 +214,32 @@ impl Socket {
...
+        self.connections.count()
+    }
+
+    #[cfg(test)]
+    pub fn forget_all_incoming_packets(&mut self) {
```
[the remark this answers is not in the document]
BourgondAries: I've changed them to private functions, but I had to change count in `connection.rs` to `pub(crate)` because it was accessed.

src/net/socket.rs
BourgondAries: Because it requires `&self` (`Socket`). I've changed it now so it's private without the cfg.

src/net/socket.rs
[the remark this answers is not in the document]
BourgondAries: Nope can't, those methods are never used if not public, so they need the cfg(test) tag to prevent clippy complaints.

src/packet.rs
jstnlef: Perhaps my biggest complaint is that it isn't a great name. It's more of a MessageId than anything.

jstnlef: Also the type names are are verbs? I don't know. I don't love it. An option would do just fine here

BourgondAries: We use item_identifier in the code to represent the item Id. I really prefer to use a separate type instead of an Option. An option is also a weaker type, we can accidentally mix two arguments wrongly.

For the verb part, it's because they're just used locally to store values, and tell the sequencer if we want a new value or use an existing value. They can be changed though. What do you propose?

jstnlef: If you really want to be that pedantic about types, technically there should then be a type alias for ItemIdentifier since it is not technically a sequence number. Then you can call it OptionItemIdentifier with Some and None varients as a new enum. I really think an option connotation is what we want here.

BourgondAries:
The ItemIdentifier has been changed to `Option<SequenceNumber>`. I would like to suggest that if we add more Options in the future, we should start to make the types more strict to avoid argument mixup.

jstnlef:
I can agree to that. That said, we should generally avoid functions/methods which take multiple (or any) optional arguments because it usually means the method is doing too many things. That said, if the generic types are different you'll still get accurate compile time errors if your options are in the wrong order.

BourgondAries:
Is jenkins correct? It's complaining about the windows duration-based test, which was fixed a while back.

TimonPost:
bors r=TimonPost,jstnlf

PR #200, opened 2019-06-16
Add method to bind to any port

BourgondAries:


src/net/socket.rs
```diff
@@ -224,6 +239,11 @@ mod tests {
     use std::thread;
     use std::time::{Duration, Instant};
 
+    #[test]
+    fn binding_to_any() {
+        Socket::bind_any().unwrap();
```
TimonPost: I would prefer to `assert!(Socket::bind_any().is_ok())`, but that is up to you.

src/net/socket.rs
```diff
@@ -38,6 +38,14 @@ impl Socket {
         Socket::bind_with_config(addresses, Config::default())
     }
 
+    /// Bind to any local port on the system, if available
+    pub fn bind_any() -> Result<(Self, Sender<Packet>, Receiver<SocketEvent>)> {
```
jstnlef: Probably want a version of this which takes a Config

BourgondAries: Done

src/net/socket.rs
```diff
@@ -38,6 +38,24 @@ impl Socket {
...
+        let loopback = Ipv4Addr::new(127, 0, 0, 1);
```
jstnlef: Now this should call `bind_any_with_config(Config::default())`

BourgondAries: Done

PR #201, opened 2019-06-16
[bugfix] Do not queue_packets twice when sequenced

BourgondAries:
Caused by a faulty if-else chain, the sequenced if would run, then the
ordered if would not run, causing its else to run, so packets became
queued twice.


BourgondAries:
Note: this also fixes that sequenced ordering simply doesn't work right now, since it will always be queued on the receiver.

BourgondAries:
The codecov checks seem to just be panics and paths we don't want taken in tests. Can we merge this anyway?

TimonPost:
bors r=TimonPost,jstnlf

Issue #202, opened 2019-06-16
Random UDP package DoS network system


o01eg:
I've tried to open UDP socket and send something here via `nc` and got panic in the networking thread:

```
thread '<unnamed>' panicked at 'not yet implemented: Packet ID 49 does not exist yet.', /home/o01eg/.cargo/registry/src/github.com-1ecc6299db9ec823/laminar-0.2.3/src/packet/enums.rs:90:18
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:39
   1: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:71
             at src/libstd/sys_common/backtrace.rs:59
   2: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:197
   3: std::panicking::default_hook
             at src/libstd/panicking.rs:211
   4: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:474
   5: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:381
   6: std::panicking::begin_panic_fmt
             at src/libstd/panicking.rs:336
   7: <laminar::packet::enums::PacketType as laminar::packet::EnumConverter>::from_u8
             at /home/o01eg/.cargo/registry/src/github.com-1ecc6299db9ec823/laminar-0.2.3/src/packet/enums.rs:90
   8: <laminar::packet::header::standard_header::StandardHeader as laminar::packet::header::header_reader::HeaderReader>::read
             at /home/o01eg/.cargo/registry/src/github.com-1ecc6299db9ec823/laminar-0.2.3/src/packet/header/standard_header.rs:99
   9: laminar::packet::packet_reader::PacketReader::read_standard_header
             at /home/o01eg/.cargo/registry/src/github.com-1ecc6299db9ec823/laminar-0.2.3/src/packet/packet_reader.rs:36
  10: laminar::net::virtual_connection::VirtualConnection::process_incoming
             at /home/o01eg/.cargo/registry/src/github.com-1ecc6299db9ec823/laminar-0.2.3/src/net/virtual_connection.rs:195
  11: laminar::net::socket::Socket::recv_from
[...]

PR #204, opened 2019-06-17
[bugfix] Prevent panic on arbitrary headers

BourgondAries:
https://github.com/amethyst/laminar/issues/202

### DIFF
--- a/src/packet/enums.rs
+++ b/src/packet/enums.rs
@@ -1,4 +1,8 @@
-use crate::packet::EnumConverter;
+use crate::{
+    error::{DecodingErrorKind, ErrorKind},
+    packet::EnumConverter,
+};
+use std::convert::TryFrom;
 
 /// Enum to specify how a packet should be delivered.
 #[derive(Copy, Clone, Debug, PartialOrd, PartialEq, Eq)]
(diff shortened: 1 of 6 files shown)


src/packet/enums.rs
```diff
@@ -55,14 +64,19 @@ impl EnumConverter for OrderingGuarantee {
             OrderingGuarantee::Ordered(_) => 2,
         }
     }
+}
 
+impl TryFrom<u8> for OrderingGuarantee {
```
TimonPost: I like this one

src/packet/enums.rs
```diff
@@ -16,13 +20,18 @@ impl EnumConverter for DeliveryGuarantee {
     fn to_u8(&self) -> u8 {
         *self as u8
     }
+}
 
+impl TryFrom<u8> for DeliveryGuarantee {
```
TimonPost: I like this one as well

src/packet/enums.rs
```diff
@@ -82,12 +96,15 @@ impl EnumConverter for PacketType {
     fn to_u8(&self) -> u8 {
```
TimonPost: maybe we can have `Into` trait implemented here instead, just a thought for later PR.

TimonPost:
bors r=TimonPost,jstnlf

PR #205, opened 2019-06-17
Use duration for the link conditioner latency

BourgondAries:


Cargo.toml
```diff
@@ -29,7 +29,8 @@ crossbeam-channel = "0.3"
 lazy_static = "1.1"
 log = "0.4"
 mio = "0.6"
-rand = "0.5"
+rand = "0.6"
+rand_pcg = "0.1"
```
TimonPost: I don't like extra deps, we should make the link container available under a certain flag eventually. Then those deps can be that as well.

src/net/socket.rs
```diff
@@ -462,4 +468,73 @@ mod tests {
             .collect();
         assert_eq!(sent_events, vec![0, 1, 35]);
     }
+
+    #[test]
+    fn really_bad_network_keeps_chugging_along() {
```
TimonPost: This starts to look like an integration test. Maybe we want to move those to /tests, I am fine to have it here as well. But just a thought.

jstnlef:
erm.. I'm curious why this was merged. 1. I didn't get a chance to review it. 2. `really_bad_network_keeps_chugging_along` test fails about 50% of the time for me on master.

Issue #195, opened 2019-06-14
Suggestion: Put Sender and Receiver inside Socket


Ploppz:
Then let all functions of `Sender` and `Receiver` such as `send()` and `recv()` be function of `Socket`. This should simplify the interface, and make the user not have to pass around 3 different values to use the socket.

BourgondAries:
This can be implemented but will likely be a breaking change... Ideally remove the looping thing altogether and let the user loop if he so desires.

TimonPost:
Removing the looping is an option, the downside of that is that we need to have changes to the API when we do want to use MIO again.

Ploppz:
I don't see how looping is relevant to this issue.

BourgondAries:
@Ploppz, we can potentially also make it even faster by making `manual_poll` use a queue instead of a thread-safe queue.

jstnlef:
I think packaging the return values into a single socket(like) wrapper struct would improve the ergonomics quite a bit.

BourgondAries:
This can be closed.

PR #209, opened 2019-06-20
Move the channels into Socket

BourgondAries:
The library should be agnostic to channels and multiple threads, meaning
that we give the user the ability to use it single-threadedly, or to
spawn threads which define their own queues that other threads can use.

This patch is the first step in that direction.

We do lose a blocking `recv` right now, but I think we can do that in a
smarter way, instead of `manual_poll` doing recv, we can let `recv` do a
direct receive on the socket instead.

Relevant issue: https://github.com/amethyst/laminar/issues/195

### DIFF
--- a/examples/server_client.rs
+++ b/examples/server_client.rs
@@ -11,7 +11,7 @@ const SERVER: &str = "127.0.0.1:12351";
 
 fn server() -> Result<(), ErrorKind> {
     let mut socket = Socket::bind(SERVER)?;
-    let (mut sender, mut receiver) = (socket.get_sender(), socket.get_receiver());
+    let (mut sender, mut receiver) = (socket.get_packet_sender(), socket.get_event_receiver());
     let _thread = thread::spawn(move || socket.start_polling());
 
     loop {
--- a/examples/server_client.rs
+++ b/examples/server_client.rs
@@ -2,45 +2,43 @@
 //! Technically, they both work the same.
 //! Note that in practice you don't want to implement a chat client using UDP.
 use std::io::stdin;
+use std::thread;
+use std::time::Instant;
 
 use laminar::{ErrorKind, Packet, Socket, SocketEvent};
-use std::thread;
 
 const SERVER: &str = "127.0.0.1:12351";
 
-fn test() {}
-
 fn server() -> Result<(), ErrorKind> {
-    let (mut socket, packet_sender, event_receiver) = Socket::bind(SERVER)?;
+    let mut socket = Socket::bind(SERVER)?;
+    let (mut sender, mut receiver) = (socket.get_sender(), socket.get_receiver());
     let _thread = thread::spawn(move || socket.start_polling());
 
-    println!("Listening for connections to {}", SERVER);
-
     loop {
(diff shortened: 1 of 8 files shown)


src/net/socket.rs
```diff
@@ -60,34 +59,40 @@ impl Socket {
...
+    /// Send a packet
+    pub fn send(&mut self, packet: Packet) {
+        let _ = self.sender.send(packet);
+    }
+
+    /// Receive a packet
+    pub fn recv(&mut self) -> Option<SocketEvent> {
+        match self.receiver.try_recv() {
+            Ok(pkt) => Some(pkt),
+            Err(_) => None,
```
jstnlef: Same as above. Probably needs to return a `Result<SocketEvent>` or `Result<Option<SocketEvent>>`

src/net/socket.rs
```diff
@@ -60,34 +59,40 @@ impl Socket {
...
+            packet_receiver,
+
+            sender: packet_sender,
+            receiver: event_receiver,
+        })
+    }
+
+    /// Send a packet
+    pub fn send(&mut self, packet: Packet) {
+        let _ = self.sender.send(packet);
```
jstnlef: Probably shouldn't swallow the error here. Would be worth transforming it into a Laminar specific error.

BourgondAries:
Ping for approval @TimonPost

src/net/socket.rs
```diff
@@ -60,34 +59,62 @@ impl Socket {
...
+            link_conditioner: None,
+            event_sender,
+            packet_receiver,
+
+            sender: packet_sender,
+            receiver: event_receiver,
+        })
+    }
+
+    /// Get a sender queue
```
jstnlef: Technically these are handles to the internal queuing mechanism.

examples/server_client.rs
```diff
@@ -2,45 +2,43 @@
...
 use laminar::{ErrorKind, Packet, Socket, SocketEvent};
-use std::thread;
 
 const SERVER: &str = "127.0.0.1:12351";
 
 fn test() {}
 
 fn server() -> Result<(), ErrorKind> {
-    let (mut socket, packet_sender, event_receiver) = Socket::bind(SERVER)?;
-    let _thread = thread::spawn(move || socket.start_polling());
```
jstnlef: We do probably want to keep at least one example that uses the polling loop.

BourgondAries: Done

src/net/socket.rs
BourgondAries: What do you propose?

jstnlef:
```
/// Returns a handle to the packet sender which provides a thread-safe way to enqueue packets to be processed. This should be used when the socket is busy running its polling loop in a separate thread. 
pub fn get_packet_sender(&mut self) -> Sender<Packet>

/// Returns a handle to the event receiver which provides a thread-safe way to retrieve events from the socket. This should be used when the socket is busy running its polling loop in a separate thread.
pub fn get_event_receiver(&mut self) -> Receiver<SocketEvent>
```

BourgondAries: Ok done

Issue #186, opened 2019-05-22
Excessive CPU usage in "start_polling" method


NGgmnBQGwR:
Using the following:

```
>rustup show
Default host: x86_64-pc-windows-gnu

stable-x86_64-pc-windows-gnu (default)
rustc 1.34.2 (6c2484dc3 2019-05-13)
```

```
>rustc -V
rustc 1.34.2 (6c2484dc3 2019-05-13)
```

```
>git rev-parse HEAD
38e129ae39ab001bad48083e3e0e98a89326fc1a
```

```
>cargo run --release --example server_client
    Finished release [optimized] target(s) in 0.10s
     Running `target\release\examples\server_client.exe`
Please type in `server` or `client`.
s
```

After which the polling thread takes the entire CPU core to himself:
![image](https://user-images.githubusercontent.com/4226721/58190729-53feae00-7ce7-11e9-9c8b-dfe461491002.png)

Which, to my limited understanding of networking, should not happen.

And possibly related - the "socket_polling_timeout" field in Config struct isn't used anywhere since its only usage was removed from "start_polling" in f2aa9e60405ad27937c71a84918f04272aa52805.

jstnlef:
I admit that the polling loop is a bit aggressive currently. Originally, to support nonblocking reads and writes to the underlying socket, we were using mio. However, we noticed that performance of mio on windows wasn't where we'd like it to be so we swapped it out to use a standard library socket in nonblocking mode. There's work to be done here to make it less abusive.

BourgondAries:
One could provide a duration to sleep between each loop.

TimonPost:
Related #191

TimonPost:
Now we have a `manual_poll ` function, I believe this issue can be closed.

PR #210, opened 2019-06-22
Add a small, customizable sleep to the polling loop

jstnlef:
Addresses https://github.com/amethyst/laminar/issues/186. Basically, we were just spinning excessively where it wasn't needed.

### DIFF
--- a/src/net/socket.rs
+++ b/src/net/socket.rs
@@ -10,7 +10,8 @@ use log::error;
 use std::{
     self, io,
     net::{Ipv4Addr, SocketAddr, SocketAddrV4, ToSocketAddrs, UdpSocket},
-    time::Instant,
+    thread::sleep,
+    time::{Duration, Instant},
 };
 
 /// A reliable UDP socket implementation with configurable reliability and ordering guarantees.
@@ -114,9 +115,16 @@ impl Socket {
     /// Entry point to the run loop. This should run in a spawned thread since calls to `poll.poll`
     /// are blocking.
     pub fn start_polling(&mut self) {
+        self.start_polling_with_duration(Duration::from_millis(1))
+    }
+
+    /// Run the polling loop with a specified duration. This should run in a spawned thread since
+    /// calls to `poll.poll` are blocking.
+    pub fn start_polling_with_duration(&mut self, poll_duration: Duration) {
         // Nothing should break out of this loop!
         loop {
             self.manual_poll(Instant::now());
+            sleep(poll_duration);
         }
     }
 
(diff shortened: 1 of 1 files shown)


jstnlef:
Seems like the coverage tool is totally busted.

fhaynes:
Out of curiosity, why sleep and not yield?

jstnlef:
A yield will merely cede control back to the CPU scheduler, which promptly re-enters the hot loop. So it goes from 100% to ~95%. A 1 ms sleep brings it down to a much more reasonable ~4%

fhaynes:
Well, it only promptly re-enters it because it has nothing else to do on the system. I would be curious to see how sleep performs on a heavily-loaded server running VMs. My guess is sleep will gradually take longer and longer than 1ms.

It would be nice to make this configurable with the sleep as the default, so that if people want to ensure performance and tight integration with the OS scheduler, they can.

jstnlef:
I hear what you're saying, but the thing is, we really don't need to be pegging the CPU. At no point do we need to be polling so often that the CPU is spinning.

fhaynes:
Pulling the packets in a way dependent on the time-keeping resolution of the OS is very different than pulling them based on the OS scheduler. As the system becomes busier, the CPU % occupied by the polling loop will decrease to smoothly accommodate other system processes.

When you make it determined by the sleep call, the OS has to do additional bookkeeping, and if it is heavily loaded, it will usually end up sleeping for longer than 1ms. Because the OS cannot even schedule it until it has decided to wake it.
What you meant to be a 1ms sleep may end up being an 8ms sleep one iteration, then 15, then back to 1.
When your OS scheduler is actually in a guest VM process that is itself being scheduled by the hypervisor, it is even more complicated, because your timer resolution can be drastically affected by VMs that have nothing to do with yours. This is a not-uncommon problem with DB clusters and other systems that require high-resolution time-keeping in AWS, etc.

jstnlef:
Made it optional 👍

src/net/socket.rs
```diff
@@ -91,11 +92,21 @@ impl Socket {
     }
 
     /// Entry point to the run loop. This should run in a spawned thread since calls to `poll.poll`
-    /// are blocking.
+    /// are blocking. We will default this to sleeping for 1ms.
     pub fn start_polling(&mut self) {
+        self.start_polling_with_duration(Some(Duration::from_millis(1)))
```
TimonPost: hardcoded value, alert, we might want to move that into the config. I know the user is able to call  `start_polling_with_duration`,  so I don't have a problem with it.

jstnlef: I really wish I could do this in rust... (sigh)
```
pub fn start_polling(&mut self, poll_duration: Option<Duration> = Some(Duration::from_millis(1))) {
```

PR #217, opened 2019-07-04
Remove mut requirement to methods which do not need it

jstnlef:
When I was integrating Laminar in our demo project, I noticed that I needed write access to the resource holding the socket when sending and receiving which seemed odd to me. Indeed, it was, in fact, odd because these methods don't require mutable access to the socket.

### DIFF
--- a/examples/udp.rs
+++ b/examples/udp.rs
@@ -22,7 +22,7 @@ fn server_address() -> SocketAddr {
 /// This is an example of how to send data to an specific address.
 pub fn send_data() -> Result<()> {
     // Setup a udp socket and bind it to the client address.
-    let mut socket = Socket::bind(client_address()).unwrap();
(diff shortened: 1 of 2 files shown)


BourgondAries:
I'm not sure about removing `&mut` from `send` and `recv`. We can envision making these mutate in the future (in the patch where channels aren't used internally) and making them just `&` would be a breaking change then.

jstnlef:
@fhaynes and @TimonPost. We really need to decide whether or not we want to keep the polling loop and the internal channels. I, for one, prefer to keep it to allow us to keep all of the stateful operations in `manual_poll` which allow us to require only read access for sends and recvs. @BourgondAries prefers that send and recv work immediately (by moving all of the socket state mutating operations into send and recv) and remove the manual poll all together. I don't think we want to live in a world where we provide both because of the complexities around that. What do you guys think?

PR #221, opened 2019-07-14
Allow a Socket to be in blocking mode

BourgondAries:
This is a useful feature in testing. We've noticed from CI that there
are timing inconsistencies even with sequential non-blocking UDP
sockets. By blocking, we can force the OS to wait for the packet to be
available.

Exposing this in the public API makes this functionality also usable by
users of the library.

`Empty` was renamed `MaybeEmpty`, because it now means 2 things: The
kernel buffer is empty (in non-blocking mode this happens), or we simply
only want to retrieve 1 packet at a time.

I'm not entirely sure if we perhaps should use another enum entry here
or kind-of overload the name. Let me know if there are better
alternatives.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -4,6 +4,8 @@ use std::{default::Default, time::Duration};
 #[derive(Clone, Debug)]
 /// Contains the configuration options to configure laminar for special use-cases.
 pub struct Config {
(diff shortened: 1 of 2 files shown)


TimonPost:
bors r=TimonPost,jstnlef

PR #223, opened 2019-08-01
Fix and extend packet reliability types documentation

futile:
I just recently started taking a closer look at laminar, and found the overview table of reliability types really really helpful!

However, I noticed 1-2 oddities and found it hard to understand what "sequenced" guarantees are from just the table. Therefore I fixed the oddities and adjusted the text a little bit.

I think this table/section is really useful for people starting out with laminar, so it makes sense to me to make this part as easy to understand as possible.

#### Fixes:
* `unreliable_sequenced`: The text says duplication is not possible, but table says
"Yes";
  * Changed this to the table also saying "No", which seems correct to me
* The row for "Reliable Sequenced" was missing from the overview table
  * Added it to the table

#### Adjustments:
* Changed possible values for "Packet Drop" from 'Yes','No' to
'Any','Any+old','Only old'.
  * This allows clearer differentiation between the different guarantees, esp. sequenced
* Changed possible values for "Packet Delivery" form 'Yes','No' to
'Yes','No','Only newest'.
  * This is a little bit more clear for sequenced guarantees as well
* Sorted the table in the same order the construcors are sorted
* Small wording adjustment after the overview table
* Realigned all tables so they look nice and clean in the code

Thanks for the great work! :)

### DIFF
--- a/src/packet/packet_structure.rs
+++ b/src/packet/packet_structure.rs
@@ -5,14 +5,15 @@ use std::net::SocketAddr;
 /// This is a user friendly packet containing the payload, endpoint, and reliability guarantees.
 /// A packet could have reliability guarantees to specify how it should be delivered and processed.
 ///
-/// | Reliability Type                 | Packet Drop | Packet Duplication | Packet Order  | Packet Fragmentation |Packet Delivery|
-/// | :-------------:                  | :-------------: | :-------------:    | :-------------:  | :-------------:    | :-------------:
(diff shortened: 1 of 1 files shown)


futile:
No idea why tests are failing, I only changed doc strings (no embedde code examples).

jstnlef:
Looks good to me. I kicked the tests. Seems like we have a transient error on windows @TimonPost

futile:
Thanks! :)

TimonPost:
bors r=jstfnl, TimonPost

PR #224, opened 2019-08-04
Implement heartbeat functionality

futile:
Make sure that connections don't time out if we don't want them to, by
regularly sending special, empty "heartbeat" packets on connections that we
have not sent on for some time.

These heartbeats have the new packet type `PacketType::Heartbeat`, which makes it possible to differentiate them from empty packets sent by a user.

The heartbeat interval can be configured and disabled/enabled via `Config::heartbeat_interval`.

~~Note: Currently these heartbeats can not be explicity disabled (but you can set a very long heartbeat interval), but that should not be hard to add, if desired.~~ Edit: Heartbeats can now be disabled.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -8,6 +8,9 @@ pub struct Config {
     pub blocking_mode: bool,
     /// Value which can specify the amount of time that can pass without hearing from a client before considering them disconnected
     pub idle_connection_timeout: Duration,
+    /// Value which specifies at which interval (if at all) a heartbeat should be sent, if no other packet was sent in the meantime.
+    /// If None, no heartbeats will be sent (the default).
+    pub heartbeat_interval: Option<Duration>,
     /// Value which can specify the maximum size a packet can be in bytes. This value is inclusive of fragmenting; if a packet is fragmented, the total size of the fragments cannot exceed this value.
     ///
     /// Recommended value: 16384
@@ -52,6 +55,7 @@ impl Default for Config {
         Self {
             blocking_mode: false,
             idle_connection_timeout: Duration::from_secs(5),
+            heartbeat_interval: None,
             max_packet_size: (MAX_FRAGMENTS_DEFAULT * FRAGMENT_SIZE_DEFAULT) as usize,
             max_fragments: MAX_FRAGMENTS_DEFAULT as u8,
             fragment_size: FRAGMENT_SIZE_DEFAULT,
--- a/src/net/socket.rs
+++ b/src/net/socket.rs
@@ -156,9 +156,19 @@ impl Socket {
             }
         }
 
-        // Finally check for idle clients
+        // Check for idle clients
         if let Err(e) = self.handle_idle_clients(time) {
             error!("Encountered an error when sending TimeoutEvent: {:?}", e);
+        }
+
+        // Finally send heartbeat packets to connections that require them, if enabled
+        if let Some(heartbeat_interval) = self.config.heartbeat_interval {
+            if let Err(e) = self.send_heartbeat_packets(heartbeat_interval, time) {
+                match e {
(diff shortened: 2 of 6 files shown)


futile:
If I understand the codecov report correctly, everything except for errors that can be returned from `Socket::send_to` is covered. Any ideas on how to test these remaining cases?

src/config.rs
```diff
@@ -52,6 +54,7 @@ impl Default for Config {
         Self {
             blocking_mode: false,
             idle_connection_timeout: Duration::from_secs(5),
+            heartbeat_interval: Duration::from_secs(2),
```
TimonPost: I think we should probably, in the future, have this calculated based on the current network circumstances. With two seconds this is safe, we might as well move it up to 3, because the idling time is 5 seconds, then the heartbeat has a total time of 2 seconds to arrive at the client.

src/net/socket.rs
```diff
@@ -156,10 +156,18 @@ impl Socket {
...
         }
 
-        // Finally check for idle clients
+        // Check for idle clients
         if let Err(e) = self.handle_idle_clients(time) {
             error!("Encountered an error when sending TimeoutEvent: {:?}", e);
         }
+
+        // Finally send heartbeat packets to connections that require them
+        if let Err(e) = self.send_heartbeat_packets(time) {
```
TimonPost: As discussed, I prefer to have this optional.

TimonPost:
Thanks for the PR, I do have one question are the acknowledgments also sent when the user only sents unreliable packets, because I don't see much-added value to have a heartbeat for those guarantees.

src/config.rs
futile: Yes, I thought about that as well, but didn't want to add RTT calculations into it (for now). Although I agree that that is probably a good idea here.

I chose 2 seconds because then if one heartbeat packet is lost, we have enough time (probably) to send another one before the connection times out. Then again, this really depends on the `idle_connection_timeout` value of the receiver, so the user currently has to adjust this value to their wishes anyway.

src/net/socket.rs
futile: Sure, should it be enabled or disabled by default?

src/config.rs
TimonPost: Right, I can understand that I am not too certain about our RTT calculations as well, so I am fine with having this statically typed now. And figure out a way to do it based on the current internet speed in the future.

src/net/socket.rs
futile: Made heartbeats disabled by default, and the config value is now an `Option<Duration>`, where `None` disables and `Some(Duration)` enables heartbeats.

futile:
I think the CI failure is spurious. I'm also working from windows and saw it only once for no reason. Also I think something similar happened during my previous documentation-PR?

futile:
> Thanks for the PR, I do have one question are the acknowledgments also sent when the user only sents unreliable packets, because I don't see much-added value to have a heartbeat for those guarantees.

I'm not quite sure what you mean, but heartbeats are currently send as unreliable and unordered, so there are no acknowledgments sent out for heartbeats atm. Did I understand your question correctly?

TimonPost:
I meant more in the way of, when sending packets, we can do use unreliably unordered to so for example. For unreliable packets, we do not have any internal state as we have for reliable ordered. So we might not care about the client to disconnect, on the other hand, I do think now, that you might also want to keep track of connected clients when only sending unreliable packets. So is not a direct problem, as long the heartbeat is optional im OK with it

jstnlef:
@TimonPost I believe that maintaining connection is an orthogonal concern to whether or not reliable packets are sent and having the option to ensure heartbeats are sent when a client is idling seems reasonable to me.

futile:
Thanks for the review! :)

TimonPost:
Agreed, let's merge this!

TimonPost:
bors r=TimonPost, jstnlef

PR #227, opened 2019-08-28
Fix ordering arranging handler for >65536 packets

BourgondAries:
The arranging handler would not reset the value of the expecting index back to 0 which caused the case of having over 2**16 packets to simply be ignored.

### DIFF
--- a/src/infrastructure/arranging/ordering.rs
+++ b/src/infrastructure/arranging/ordering.rs
@@ -244,6 +244,7 @@ impl<T> Arranging for OrderingStream<T> {
     ) -> Option<Self::ArrangingItem> {
         if incoming_offset == self.expected_index {
             self.expected_index += 1;
+            self.expected_index %= u16::max_value() as usize + 1;
             Some(item)
         } else if incoming_offset > self.expected_index {
             self.storage.insert(incoming_offset, item);
--- a/src/infrastructure/arranging/ordering.rs
+++ b/src/infrastructure/arranging/ordering.rs
@@ -234,7 +234,7 @@ impl<T> Arranging for OrderingStream<T> {
     /// This can only happen in cases where we have a duplicated package. Again we don't give anything back.
     ///
     /// # Remark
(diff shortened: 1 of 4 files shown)


BourgondAries:
@TimonPost looks like it, but I'd like to cover it in a `socket.rs`-level test case first.

src/infrastructure/arranging/ordering.rs
```diff
@@ -244,6 +244,7 @@ impl<T> Arranging for OrderingStream<T> {
     ) -> Option<Self::ArrangingItem> {
         if incoming_offset == self.expected_index {
             self.expected_index += 1;
+            self.expected_index %= u16::max_value() as usize + 1;
```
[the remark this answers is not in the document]
BourgondAries: No socket.rs test but an ordering.rs test instead!

src/infrastructure/arranging/ordering.rs
```diff
@@ -234,18 +234,25 @@ impl<T> Arranging for OrderingStream<T> {
...
         incoming_offset: usize,
         item: Self::ArrangingItem,
     ) -> Option<Self::ArrangingItem> {
+        let incoming_offset = incoming_offset as u16;
         if incoming_offset == self.expected_index {
-            self.expected_index += 1;
+            self.expected_index = self.expected_index.wrapping_add(1);
             Some(item)
-        } else if incoming_offset > self.expected_index {
+        } else if (self.expected_index < u16::max_value() / 2
```
TimonPost: this is a bit complex, maybe it will be helpful if there is a small comment saying what the purpose of it is.

BourgondAries: It is, maybe there's something that exists so we can do `self.expected_index.within_next(u16::max_value(), incoming_offset)`.

BourgondAries: Should be resolved with the new function

BourgondAries:
Let me reopen a PR because this one doesn't appear to track my fork for some reason.

PR #228, opened 2019-08-31
Turn arranging usize into u16

BourgondAries:
Since the protocol specifies the use of 16 bit counters, it's best to
reflect these in the API as well.


TimonPost:
This is create let's merge this

PR #229, opened 2019-09-01
Fix ordering arranging handler for >65536 packets

BourgondAries:
The arranging handler would not reset the value of the expecting index back to 0 which caused the case of having over 2**16 packets to simply be ignored.

Note: This is https://github.com/amethyst/laminar/pull/227 but for some reason it didn't track my commits after merging https://github.com/amethyst/laminar/pull/228, so that one got closed and this one opened. (Apparently had upstreamed to master...)

### DIFF
--- a/src/infrastructure/arranging/ordering.rs
+++ b/src/infrastructure/arranging/ordering.rs
@@ -244,6 +244,7 @@ impl<T> Arranging for OrderingStream<T> {
     ) -> Option<Self::ArrangingItem> {
         if incoming_offset == self.expected_index {
             self.expected_index += 1;
+            self.expected_index %= u16::max_value() as usize + 1;
             Some(item)
         } else if incoming_offset > self.expected_index {
             self.storage.insert(incoming_offset, item);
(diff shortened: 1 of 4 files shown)


src/infrastructure/arranging/ordering.rs
```diff
@@ -234,18 +243,19 @@ impl<T> Arranging for OrderingStream<T> {
...
-    /// - When we receive an item there is a possibility that a gab is filled and one or more items will could be returned.
+    /// - When we receive an item there is a possibility that a gap is filled and one or more items will could be returned.
     ///   You should use the `iter_mut` instead for reading the items in order.
     ///   However the item given to `arrange` will be returned directly when it matches the `expected_index`.
     fn arrange(
         &mut self,
         incoming_offset: u16,
         item: Self::ArrangingItem,
     ) -> Option<Self::ArrangingItem> {
+        let incoming_offset = incoming_offset as u16;
```
jstnlef: This value is already u16 no?

BourgondAries: Yes, missed that during rebasing from the previous patch.

PR #233, opened 2019-09-01
Ensure that `self.remote_ack_sequence_num` is always increasing

jstnlef:
We don't want out of order delivery of reliable packets to set the property to a value `sequence_less_than` its current value. However, we do still want to process the packet since it will contain acks about older packets we should remove from the sent packets list.


BourgondAries:
Might want to test with some packets crossing the u16 wrap boundary.

PR #234, opened 2019-09-02
Disconnect the connection after sending N un-acked packets

BourgondAries:
The connection is disconnected if we have N packets-in-flight
simultaneously. Under normal usage we expect packets to be acked
regularly so that our packets-in-flight size is relatively small.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -48,6 +48,12 @@ pub struct Config {
     ///
     /// Value that specifies how long we should block polling for socket events, in milliseconds. Defaults to `1ms`.
     pub socket_polling_timeout: Option<Duration>,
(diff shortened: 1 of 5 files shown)


BourgondAries:
@jstnlef What do you propose? This value is basically a "disable" value as it stands.

jstnlef:
What about 512? Less than 20 seconds at 30 Hz send rate.

BourgondAries:
Sounds good

azriel91:
Had to abort the build (sorry -- the build agent got stuck and had to be restarted). I think you have to push another commit to restart a build.

BourgondAries:
Codecov is just complaining about panics not being triggered in a test.

PR #237, opened 2019-09-06
Codecov action

TimonPost:


TimonPost:
I can now experiment with my own crate, going to close this, since we don't want multiple CI's, as spoken about in discord.

PR #243, opened 2019-09-09
Dependency maintenance

TimonPost:
Haha, we even had mio as dep


BourgondAries:
We should probably add `cargo udeps` to the CI

Issue #242, opened 2019-09-09
More informative connection states and ability to manage them


fraillt:
hello,
I would like to contribute to Laminar, basically, I want to do several things (mostly in `net::socket.rs`):
1) get all events (and more informative) related to connection state:
* created(addr) - when the connection is actually added to active connections list
* connected(addr) - when connection receive first incoming packet
* packet(addr, data) - actual received packet data, (this is the only event that can happen multiple times)
* disconnected(addr, reason) - when we start to shut-down connection and no longer send or receive packets. It could be multiple reasons, e.g.: 
	* disconnect timeout, (error: destroys immediately)
	* too many packets in flight, (error: destroys immediately)
	* close request from the sender (sender may ask to finish sending reliable packets). This also should provide duration, when the connection should be actually destroyed. This will allow the most efficient way to control DDoS attacks, because all packets will be rejected in the laminar layer, without propagating them to the application layer.
	* destroy request from the sender.
* destroyed(addr) - when the connection is actually removed from active connections.
2) change `Sender<Packet>` to `Sender<SocketSendEvent>` where `SocketSendEvent` will be with following following enum :
* open(addr) - send request to actually add active connection
* packet(addr, data) - send actual packet data
* close(addr, duration, finish_in_flight) - initiate disconnect and provide duration for how long this connection should ignore incoming requests, this will allow to control DDoS or prevent from creating the connection again when old packets from the client might recreate connection again. A Sender also has the option to finish sending current reliable packets or not.
* destroy(addr) - (rarely used) this will force to remove the active connection from the list immediately. This is the only way to directly remove the connection from the list, that might be useful in cases where you want reenable connections that are still not destroyed due to close duration not expired.

I believe that this will simplify usage for everyone and solve multiple issues:
* This will prevent cases where connections are created without sending an event, e.g. multiple Timeout, or Connected events can be received in a row.
This implementation ensures that all connections will go through these states: created(1)->connected(0..1)->packet(0..N)->disconnected(1)->destroyed(1).
* Allow to easily track the exact state of all connections in laminar, because currently some events are created implicitly and it is impossible to know what is exactly happening inside,
* Also provides most efficient DDoS prevention, current DDoS prevention implementation is not optimal, because laminar needs to process all packets and all events*2 (Connect + Packet) are sent to the application, where instead they should all be ignored. There should be some sort of DDoS monitoring module, that will alert when ignored packets reach a certain threshold.
[...]

Issue #153, opened 2019-03-10
Implement metrics.


TimonPost:
We should have metrics for laminar to make debugging easier.

1. `SocketEvent` should have an entry for `Metrics`:
```
pub enum SocketEvent {
    /// A packet has been received from a client.
    Packet(Packet),
    /// A new client connects. Clients are uniquely identified by the ip:port combination at this layer.
    Connect(SocketAddr),
    /// A client disconnects. This is generated from the server-side intentionally disconnecting a client,
    /// or it could be from the client disconnecting.
    Disconnect(SocketAddr),
    /// This is generated if the server has not seen traffic from a client after a configurable amount of time.
    Timeout(SocketAddr),
    
    /// Metrics of lamianar
    Metrics(Metrics)
}
```

2. Example of metrics we can gather:
[source](https://github.com/jstnlef/mercury/blob/master/protocol/src/metrics.rs#L19)
```
pub struct Metrics {
    counters: [u64; DataPoint::Length as usize],
    packet_loss: f32,
    sent_bandwidth_kbps: f32,
    received_bandwidth_kbps: f32,
    acked_bandwidth_kbps: f32,

    // Config values to tweak the calculated fields
    bandwidth_smoothing_factor: f32,
}
````
3. Send metrics back to the user
Because we created a Metrics entry for the `SocketEvent` we can send it on the [EventChannel](https://github.com/amethyst/laminar/blob/master/src/net/socket.rs#L24)

4. Add a config entry 

- Metrics Enabled; whether we should calculate metrics or not.
- Metrics send interval; send on event channel.

**Comment by TimonPost** [comment]
We should have metrics for laminar to make debugging easier.

1. `SocketEvent` should have an entry for `Metrics`:
```
pub enum SocketEvent {
    /// A packet has been received from a client.
    Packet(Packet),
    /// A new client connects. Clients are uniquely identified by the ip:port combination at this layer.
    Connect(SocketAddr),
    /// A client disconnects. This is generated from the server-side intentionally disconnecting a client,
    /// or it could be from the client disconnecting.
    Disconnect(SocketAddr),
    /// This is generated if the server has not seen traffic from a client after a configurable amount of time.
    Timeout(SocketAddr),
    
    /// Metrics of lamianar
    Metrics(Metrics)
}
```

2. Example of metrics we can gather:
[source](https://github.com/jstnlef/mercury/blob/master/protocol/src/metrics.rs#L19)
```
pub struct Metrics {
[...]

PR #246, opened 2019-09-18
WIP connection manager

fraillt:
This PR provides a concept of a connection with connected/disconnected states over UDP.
It provides two traits `ConnectionManager` and `SocketManager`, and implementation of these can be provided by the user. More discussion about it in this issue #242.
The emphasis of this PR is a connection manager, although it also provides a socket manager concepts, after talking to @TimonPost on discord, I think it is wise to split it apart.
Currently, a socket manager basically does two things:
* tracks system errors and bytes sent/received/ignored
* based on this information, it decided if a connection can be created, and when to destroy it.

Probably it is wise to implement some metrics tracking in laminar itself, (some discussion about it is here #153.) and change `SocketManager` to `ConnectionManagerFactory` which is only responsible for creating `ConnectionManager`.
Decision making of whether to accept or reject new address can be addressed in the future.


src/net/virtual_connection.rs
```diff
@@ -38,11 +39,18 @@ pub struct VirtualConnection {
 
     config: Config,
     fragmentation: Fragmentation,
+    pub state_manager: Box<dyn ConnectionManager>,
+    pub current_state: ConnectionState,
 }
 
 impl VirtualConnection {
     /// Creates and returns a new Connection that wraps the provided socket address
```
TimonPost:
```suggestion
    /// Creates and returns a new `VirtualConnection` that wraps the provided socket address
```

src/net/virtual_connection.rs
```diff
@@ -61,6 +71,14 @@ impl VirtualConnection {
         self.acknowledge_handler.packets_in_flight() > self.config.max_packets_in_flight
     }
 
+    pub fn is_disconnected(&self) -> bool {
+        if let ConnectionState::Disconnected(_) = self.current_state {
```
TimonPost: can't we do something like: `return ConnectionState::Disconnected(_) == self.current_state`

src/net/virtual_connection.rs
```diff
@@ -75,6 +93,10 @@ impl VirtualConnection {
         time.duration_since(self.last_sent)
     }
 
+    pub fn get_current_state(&self) -> &ConnectionState {
```
TimonPost: doc string

src/packet/enums.rs
```diff
@@ -107,6 +109,7 @@ impl TryFrom<u8> for PacketType {
             0 => Ok(PacketType::Packet),
             1 => Ok(PacketType::Fragment),
             2 => Ok(PacketType::Heartbeat),
+            3 => Ok(PacketType::ConnectionManager),
```
TimonPost: `ConnectionManager` sounds odd for a packet type. Maybe we should call this just `Connection`

src/packet/packet_structure.rs
```diff
@@ -176,6 +176,58 @@ impl Packet {
     }
 }
 
+/// This packet type have similar properties to `Packet` except that it doesn't own anything, and additionally has `PacketType`.
+#[derive(Debug)]
+pub struct GenericPacket<'a> {
+    /// defines packet type, currently Fragment type should not set,
```
TimonPost:
```suggestion
    /// defines packet type, currently `Fragment` type is not supported
```

src/packet/packet_structure.rs
```diff
@@ -176,6 +176,58 @@ impl Packet {
...
+    /// the raw payload of the packet
+    pub(crate) payload: &'a [u8],
+    /// defines on how the packet will be delivered.
+    pub(crate) delivery: DeliveryGuarantee,
+    /// defines on how the packet will be ordered.
+    pub(crate) ordering: OrderingGuarantee,
+}
+
+impl<'a> GenericPacket<'a> {
+    pub fn manager_packet(
```
TimonPost: doc string

src/packet/packet_structure.rs
```diff
@@ -176,6 +176,58 @@ impl Packet {
...
+        ordering: OrderingGuarantee,
+    ) -> Self {
+        Self {
+            packet_type: PacketType::ConnectionManager,
+            payload,
+            delivery,
+            ordering,
+        }
+    }
+    pub fn user_packet(
```
TimonPost: enter
docstring

src/packet/packet_structure.rs
```diff
@@ -176,6 +176,58 @@ impl Packet {
...
+        ordering: OrderingGuarantee,
+    ) -> Self {
+        Self {
+            packet_type: PacketType::ConnectionManager,
+            payload,
+            delivery,
+            ordering,
+        }
+    }
+    pub fn heartbeat_packet(
```
TimonPost: enter
docstring

src/managers.rs
```diff
@@ -0,0 +1,4 @@
+//! This module provides socket managers.
```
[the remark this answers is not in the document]
fraillt: Would you agree that I split SocketManager into two parts, as I wrote in the comment?
One would be `ConnectionManagerFactory` two methods `accept_remote_connection`, `accept_local_connection`, and the other would be SystemTracker?, SocketTracker?, MetricsCollector? ... you name it :) that would basically have these `track_` methods.

src/managers/simple.rs
```diff
@@ -0,0 +1,206 @@
...
+use crate::packet::{DeliveryGuarantee, OrderingGuarantee};
+use log::error;
+use std::collections::VecDeque;
+use std::io::ErrorKind::WouldBlock;
+use std::net::SocketAddr;
+use std::time::Instant;
+
+/// The simplest connection manager, that immediately goes into connected state, after creating it
+#[derive(Debug, Default)]
+struct AlwaysConnectedConManager {
```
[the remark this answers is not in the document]
fraillt: Ok, sounds fair.

src/managers/simple.rs
```diff
@@ -0,0 +1,206 @@
...
+    connected: bool,
+}
+
+impl ConnectionManager for AlwaysConnectedConManager {
+    fn update<'a>(
```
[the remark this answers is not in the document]
fraillt: There is nothing to pass to it, the way this manager works, is that it returns `Some(Ok(Either::Right(ConnectionState::Connected(Box::default()))))` only on the first call to `update`, which allows to switch to connected state, and leave in this state until connection is forcibly closed.

src/managers/simple.rs
```diff
@@ -0,0 +1,206 @@
...
+use log::error;
+use std::collections::VecDeque;
+use std::io::ErrorKind::WouldBlock;
+use std::net::SocketAddr;
+use std::time::Instant;
+
+/// The simplest connection manager, that immediately goes into connected state, after creating it
+#[derive(Debug, Default)]
+struct AlwaysConnectedConManager {
+    connected: bool,
```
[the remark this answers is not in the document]
fraillt: It is used, to return initial state only on the first call to `update`.
I changed it to `initial_state: Option<ConnectionState>` and wrote a comment, that explains its purpose.

src/managers/simple.rs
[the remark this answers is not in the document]
fraillt: Not sure if we have to.
Socket manager is public to the user, and the connection manager is an implementation detail of it.
If we rename socket manager to ConnectionManagerFactory or something else, maybe we can rename /src/managers.rs, /src/managers/ as well...

src/managers/simple.rs
```diff
@@ -0,0 +1,206 @@
...
+                    self.change_state(ConnectionState::Disconnected(TargetHost::RemoteHost));
+                }
+            }
+            _ => panic!("In disconnected nothing can happen"),
+        }
+        Ok(())
+    }
+
+    fn connect(&mut self, data: Box<[u8]>) {
+        self.send_packet([b"connect-", data.as_ref()].concat().as_ref());
```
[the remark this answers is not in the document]
fraillt: Ok

src/managers/simple.rs
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    fn disconnect(&mut self) {
+        if let ConnectionState::Connected(_) = self.state {
+            self.send_packet(b"disconnect");
+        }
+        self.change_state(ConnectionState::Disconnected(TargetHost::LocalHost));
```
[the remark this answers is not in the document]
fraillt: Currently, it is only used to inform who initiated a disconnect (local or remote) host.

src/net/connection.rs
```diff
@@ -13,12 +22,15 @@ use std::{
 #[derive(Debug)]
 pub struct ActiveConnections {
     connections: HashMap<SocketAddr, VirtualConnection>,
+    buffer: Box<[u8]>,
```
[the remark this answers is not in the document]
fraillt: probably it is subproduct of refactoring ;)

src/net/connection.rs
```diff
@@ -81,16 +127,117 @@ impl ActiveConnections {
...
+                                ConnectionState::Disconnected(closed_by) => {
+                                    sender.send(ConnectionEvent(
+                                        conn.remote_address,
+                                        ReceiveEvent::Disconnected(DisconnectReason::ClosedBy(
+                                            closed_by.clone(),
+                                        )),
+                                    ))
+                                }
+                                _ => panic!(
+                                    "Invalid state transition: {:?} -> {:?}",
```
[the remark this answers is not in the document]
fraillt: This is a tricky one.
It is a connection manager error, but it is created by laminar, so in order to properly handle it I need to provide some kind of severity levels to connection manager errors so that connection manager error handling could be independent of socket manager.

I came up with this modification:
```rust
/// Generic error type, that is used by ConnectionManager implementation.
#[derive(Debug, PartialEq, Clone)]
pub enum ConnectionManagerError {
    /// Something really bad has happened, this is not a recoverable error, and the connection should be destroyed.
    Fatal(String),
    /// Something unexpected has happened, but the connection is still in a valid state.
    /// `SocketManager` can decide when to destroy the connection if two many warnings are propagated from the same connection in a short amount of time.
    Warning(String)
    // TODO: is it enought?
}
```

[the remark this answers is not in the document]
fraillt: And changed this panic! to ConnectionManagerError::Fatal

src/net/connection.rs
```diff
@@ -81,16 +127,117 @@ impl ActiveConnections {
...
+                            } {
+                                manager.track_connection_error(
+                                    &conn.remote_address,
+                                    &ErrorKind::SendError(SendError(Either::Right(err.0))),
+                                    "sending connection state update",
+                                );
+                            }
+                        } else {
+                            panic!(
+                                "Invalid state transition {:?} -> {:?}",
```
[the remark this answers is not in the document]
fraillt: changed to ConnectionManagerError::Fatal

src/net/managers.rs
```diff
@@ -0,0 +1,135 @@
...
+    /// Accepts full packet: header + payload
+    fn postprocess_outgoing<'a, 'b>(&mut self, data: &'a [u8], buffer: &'b mut [u8]) -> &'b [u8]
+    where
+        'a: 'b;
+
+    /// This will be called only for incoming protocol specific packets, after laminar's reliability layer accepted it.
+    /// This is convenient place to process actual logic, because it is filtered by laminar's reliability layer and it accepts only `PacketType::ConnectionManager` messages.
+    fn process_protocol_data(&mut self, data: &[u8]) -> Result<(), ConnectionManagerError>;
+
+    /// This will be invoked when player sends connect request,
```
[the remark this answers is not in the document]
fraillt: it's layer, not player ;) or you don't like the term "laminar's reliability layer"?

src/net/managers.rs
```diff
@@ -0,0 +1,135 @@
...
+    /// Some protocols might provide a way to pass initial connection data, hence the `data` field.
+    /// This method can only be called when connection is in `Connecting` state
+    fn connect(&mut self, data: Box<[u8]>);
+
+    /// This will be invoked when player sends SendEvent::Disconnect request.
```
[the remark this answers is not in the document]
fraillt: oops :) I'll change it to "user" then.

src/net/virtual_connection.rs
fraillt: I don't know how to do this... because I need to pattern match here. I don't want to compare for equality, because what I care is that state is Disconnected.

src/packet/enums.rs
fraillt: Ok, this sounds better

fraillt:
I extracted part of `VirtualConnection` into separate class `ReliabilitySystem` which knows how to parse from bytes to packets, and vice- versa. And result type implements `Iterator` trait, which is convenient.

I want to finish extract stuff from VirtualConnection, and leave it only simple struct with no implementation with these fields:
```rust
pub struct VirtualConnection {
    pub remote_address: SocketAddr,
    pub reliability_system: ReliabilitySystem,
    pub connection_manager: Box<dyn ConnectionManager>,
    pub current_state: ConnectionState,
}
```
and I would really like to put `last_heard` and `last_sent` into separate struct named `ActivitySystem` or something, and put it in connection manager.
Do you approve these changes?

TimonPost:
It is difficult to say if I would approve it. I do have questions and concerns. 1) This PR is quite big we don't change such big things. A lot of tests and logic is bound to virtual connection. So breaking that will break other stuff. That's where my second question comes in 2) What is your motivation behind those changes? Is it needed, are you having problems with implementing the feature. I prefer not to change anymore other than that is needed for the connection system.

fraillt:
Ye, I do agree with you, it would be best if I freeze this branch and only fix bugs and write tests. I'll remove this last commit.
But I do see this as an improvement, because of Rusts borrowing rules, this would allow for better optimizations in the future. As a side effect it is easier to understand for developer as well:) basically what I don't like at current state is that I have ConnectionManager and ConnectionState in VirtualConnection as public fields, that is not used in the class itself, but used by higher abstractions, I put them here because it was the only place to store per connection information:) My extracted new interface is basically renamed VirtualConnection:)
But lets do cleanup things latter, as separate PR

src/packet/packet_structure.rs
```diff
@@ -176,6 +176,61 @@ impl Packet {
...
+    pub(crate) payload: &'a [u8],
+    /// defines on how the packet will be delivered.
+    pub(crate) delivery: DeliveryGuarantee,
+    /// defines on how the packet will be ordered.
+    pub(crate) ordering: OrderingGuarantee,
+}
+
+impl<'a> GenericPacket<'a> {
+    /// Creates a connection manager specific packet
+    pub fn manager_packet(
```
TimonPost: manager packet sounds weird, maybe we should change that name to connection packet

src/packet/header/standard_header.rs
```diff
@@ -63,6 +63,11 @@ impl StandardHeader {
         self.packet_type == PacketType::Fragment
     }
 
+    /// Returns true if packet is only used by `ConnectionManager`
+    pub fn is_connection_manager_packet(&self) -> bool {
```
TimonPost: same here, below we had manager packet, and here we have connection manager packet.

src/packet/enums.rs
```diff
@@ -84,12 +84,14 @@ impl TryFrom<u8> for OrderingGuarantee {
 #[derive(Copy, Clone, Debug, PartialOrd, PartialEq)]
 /// Id to identify a certain packet type.
 pub enum PacketType {
-    /// Full packet that is not fragmented
+    /// User full packet that is not fragmented,
     Packet = 0,
-    /// Fragment of a full packet
+    /// User fragment of a full packet
```
TimonPost: I don't see the added value of user here

src/packet/enums.rs
```diff
@@ -84,12 +84,14 @@ impl TryFrom<u8> for OrderingGuarantee {
 #[derive(Copy, Clone, Debug, PartialOrd, PartialEq)]
 /// Id to identify a certain packet type.
 pub enum PacketType {
-    /// Full packet that is not fragmented
+    /// User full packet that is not fragmented,
```
TimonPost:
```suggestion
    /// User full packet that is not fragmented
```

src/net/virtual_connection.rs
```diff
@@ -232,11 +263,13 @@ impl VirtualConnection {
         }
     }
 
-    /// This processes the incoming data and returns a packet if the data is complete.
+    // TODO would be super nice to return iterator of new packets, instead of passing `sender`, and `socket_manager` inside.
```
TimonPost: We try to avoid TODO comments in laminar. So once this is merged, create an issue for this.

src/net/virtual_connection.rs
TimonPost: Okay, don't mind then

src/managers/simple.rs
```diff
@@ -0,0 +1,213 @@
...
+use std::io::ErrorKind::WouldBlock;
+use std::net::SocketAddr;
+use std::time::Instant;
+
+/// The simplest connection manager, that immediately goes into the connected state, after creating it
+#[derive(Debug)]
+struct AlwaysConnectedConnectionManager {
+    // this is used to set initial state as connected when creating an instance.
+    // we'll take this value on first `update` call
+    initial_state: Option<ConnectionState>,
```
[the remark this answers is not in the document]
fraillt: `take` method on `Option` moves current value into new otional, and leave the old one in the `None` state.
This is exactly what I want to achieve. Return Some(...State::Connected), and all other calls to `update` returns `None`.

src/packet/enums.rs
fraillt: I wanted to emphasize, that these are only for user packets, because connection and heartbeat packets doesn't support this.

src/error.rs
```diff
@@ -27,9 +29,11 @@ pub enum ErrorKind {
     /// Protocol versions did not match
     ProtocolVersionMismatch,
     /// Could not send on `SendChannel`.
-    SendError(SendError<SocketEvent>),
+    SendError(SendError<Either<ConnectionEvent<SendEvent>, ConnectionEvent<ReceiveEvent>>>),
```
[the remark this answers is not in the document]
fraillt: It is a send error from the point of view of crossbeam crate, and it is used for sending events and receiving events. `SocketEvent` was actually renamed to `ReceiveEvent` because I introduced new type, `SendEvent`. So the way it works, hasn't changed.
I do agree that it sounds confusing, but what should I rename to
`ChannelSenderError(SendError<Either<ConnectionEvent<SendEvent>, ConnectionEvent<ReceiveEvent>>>` ?

src/net/managers.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// This function should be called frequently, even if there are no packets to send or receive.
+    /// It will always be called last, after all, other methods are called, in the main laminar`s loop.
+    /// It can generate all kinds of packets: heartbeat, user or connection protocol packets.
+    /// (maybe heartbeat functionality should be moved here?)
+    /// It will be called in the loop as long, as it returns any results. E.g. `connect` method may generate multiple results: change state and send packet.
+    fn update<'a>(
+        &mut self,
+        buffer: &'a mut [u8],
+        time: Instant,
+    ) -> Option<Result<Either<GenericPacket<'a>, ConnectionState>, ConnectionManagerError>>;
```
jstnlef: This return value terrifies me. Really smells like this method is trying to do way too many things. There is most definitely a better structure here.

jstnlef:
I've taken a bit of time to look through this and while I appreciate the need for connection management in laminar, I'm not too keen on this implementation. This feels very "bolted" on and not at all integrated with the solution proper. For example, I would expect something called `ConnectionManager` to own (and or manage) the connections. This makes the interaction point with it in `recv_from` extremely awkward.
```
let conn = self.connections.get_or_insert_connection(
    address,
    &self.config,
    time,
    manager,
);
ActiveConnections::update_connection_manager(
    conn,
    &self.event_sender,
    self.manager.as_mut(),
    &mut self.socket,
    time,
    self.tmp_buffer.as_mut(),
);
```
Another bit of weirdness is around the caller API. Having to wrap `socket.get_event_sender()` as SocketEventSender(socket.get_event_sender()) is not exactly ergonomic nor obvious to users.

I'm also worried about the proliferation of the `Either` type as well as the increase in the size of wrapped types. e.g.
`Err(ErrorKind::SendError(SendError(Either::Left(error.0))))`
and
`Option<Result<Either<GenericPacket<'a>, ConnectionState>, ConnectionManagerError>>`

Anyway, I urge you to rethink your implementation. I really do think that this feature is necessary and want to see it be a success because I'm really digging the end result but I think that it might be worth taking some extra time to figure out how to integrate it more holistically.

Top suggestions:
1. Have the new `ConnectionManager` own the lifecycle of connections. Creation, updates, deletion, etc of `VirtualConnection`
2. Replace usages of `Either` type with enums of an appropriate name. If you can't think of a name that unifies the 2 types, perhaps you're dealing with 2 different behaviors that should be handled by different functions/methods.
3. Let's see if we can get away with fewer wrappers. Looking at you `SocketEventSender` and `ConnectionEvent`. That said, the split of `SocketEvent` into `ReceiveEvent` and `SendEvent` is good.
4. The bouncing between socket methods and `SocketManager` methods is strange. I'm not even entirely sure `SocketManager` is needed since it isn't exactly "managing" sockets. I would say we should toss it.

src/packet/enums.rs
TimonPost: Ah okay thanks for the explanation makes more sense now

src/managers/simple.rs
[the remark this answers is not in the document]
TimonPost: Ah okay, understand your intention better now.

src/error.rs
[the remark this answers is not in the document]
fraillt: changed to `ChannelSenderError(SendError<Either<ConnectionEvent<SendEvent>, ConnectionEvent<ReceiveEvent>>>)`

src/packet/header/standard_header.rs
fraillt: renamed to `is_connection_packet`

src/net/virtual_connection.rs
fraillt: I was planning to do exactly that.

src/net/managers.rs
fraillt: Good point, it doesn't look very intuitive. I updates to this:
```rust
/// Events that are generated by `ConnectionManager`s update method.
#[derive(Debug)]
pub enum ConnectionManagerEvent<'a> {
    /// Generated new packet, that will be sent immediatelly.
    NewPacket(GenericPacket<'a>),
    /// `ConnectionState` state changed.
    NewState(ConnectionState),
    /// Error occured within `ConnectionManager`
    Error(ConnectionManagerError)
}
```
and return type is now `Option<ConnectionManagerEvent<'a>>`

jstnlef: That already looks much better.

fraillt:
I'm glad you addressed these points because most of them I don't like either. But they are more or less side effects of previous discussions. I guess I was trying to address too many things in this PR.

This code, that you quoted, looks weird for me too, partially because part of the connection is constructed inside ActiveConnections and part of it in Socket implementation.
As I have mentioned earlier, I think we need to split VirtualConnection into more parts, because it is too broad term now.

`SocketEventSender` is, not well thought enough, product of discussion #242, the concern was that `client.send(ConnectionEvent(addr,SendEvent::Packet(Packet::unreliable(addr, .......))));` is too verbose, compared to `client.send(Packet::unreliable(addr, .......));` and I fully agree.
Maybe instead of providing crossbeam Sender directly through socket `get_event_sender` method, we should return a wrapper object, with only `connect`, `disconnect`, `send_packet` methods?

`SocketManager` is also product of discussion in the #242, there was a lot discussion about how DoS should be handled, and socket manager is an attempt to provide interface, that could be used to handle connection lifetime, but I do agree this is not necessary and I already expressed my concern in initial comment of this PR that maybe it should be changed to `ConnectionManagerFactory`?

> I would expect something called ConnectionManager to own (and or manage) the connections.

Maybe you have a better name to suggest?
* ConnectionStateManager - not precise enough, because it does more than one thing?
* ConnectionPostPreProcessor + ConnectionStateManager ?
* ConnectionHandler?

Last questions: why do you think `ConnectionEvent` is bad abstraction? I think it perfectly unites two heterogeneous things: what event happened, and who is the target of event. With this separation in place, events could also be reused in the application layer, where you usually have a separate object for each address already.

Ok, so until these questions are answered, I'll do the following:

1. Change a return type of `update` method in ConnectionManager: 
```rust
pub enum ConnectionManagerEvent<'a> {
    NewPacket(GenericPacket<'a>),
    NewState(ConnectionState),
    Error(ConnectionManagerError)
}
```
2. Rename VirtualConnection to ReliabilitySystem, and new VirtualConnection would look like this:
```rust
pub struct VirtualConnection {
    pub remote_address: SocketAddr,
    pub reliability_system: ReliabilitySystem,
    pub connection_manager: Box<dyn ConnectionManager>,
    pub current_state: ConnectionState,
}
```
3. Replace `Either` with more precise names or functions.
4. Leave SocketEventSender and ConnectionEvent as it is, and wait for your suggestions, because I don't know what is the best approach to address this.
5. Split SocketManager into two parts, ConnectionManagerFactory and MetricsCollector as I wrote in [here](https://github.com/amethyst/laminar/pull/246#discussion_r325988763).

So, let's get to work :)

src/error.rs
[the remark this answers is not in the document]
fraillt: Changed to 
```rust
ChannelSendingError(SendError<ConnectionEvent<SendEvent>>),
ChannelReceivingError(SendError<ConnectionEvent<ReceiveEvent>>),
```

fraillt:
This PR is getting bigger, and bigger ...
I want to point out, that `MetricsCollector` is unrelated to connection manager, and currently basically serves as a logger. I can remove it if you want, but this might be useful in future for #153 and #239.

Code could be made much cleaner if we simply ignore [SendError](https://docs.rs/crossbeam-channel/0.1.2/crossbeam_channel/struct.SendError.html) from the crossbeam. 
As I understand it can never happen anyway because `Socket` always stores sender and receiver pairs.

fraillt:
Since this PR gone out of control, in terms of what amount of code has changed. I decided to split it in smaller PR and prepare a code base for easier integration of connection manager in the future.

PR #251, opened 2019-09-22
[ordered]: Fix sending acks for ordered fragmented packets

BourgondAries:
The packets would not be acked because only the first fragment had an
ack header, and that ack header would only be used after the last
fragment had been received - while the header had not been stored.

This patch stores that header inside the reassembly data so it can later
be used when returning the final packet.


TimonPost:
So, in short, does this PR add acknowledged fragments?

BourgondAries:
No, it means that a collection of fragments is acknowledged, not the individual fragment.

PR #252, opened 2019-09-22
[ordered]: Ensure we don't read out-of-bounds on malformed headers

BourgondAries:
Malformed headers can read out of bounds if the input isn't checked.

Suggest that we fuzz sending arbitrary data to uncover more flaws.

### DIFF
--- a/src/infrastructure/fragmenter.rs
+++ b/src/infrastructure/fragmenter.rs
@@ -120,6 +120,10 @@ impl Fragmentation {
                 Err(FragmentErrorKind::FragmentWithUnevenNumberOfFragemts)?
             }
 
+            if usize::from(fragment_header.id()) >= reassembly_data.fragments_received.len() {
+                Err(FragmentErrorKind::ExceededMaxFragments)?;
+            }
+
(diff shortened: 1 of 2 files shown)


src/infrastructure/fragmenter.rs
```diff
@@ -120,6 +120,10 @@ impl Fragmentation {
                 Err(FragmentErrorKind::FragmentWithUnevenNumberOfFragemts)?
             }
 
+            if usize::from(fragment_header.id()) >= reassembly_data.fragments_received.len() {
```
TimonPost: This is what I don't understand. You have a test with only the standard header with acknowledge header. This packet has the type of 'Fragmented' and is reliable and ordered. Now I guess that you don't add the fragment header intentionally, however, this is where my confusion starts.  Here you try to access `fragment_header` which is already parsed when the code is here.

 If you don't append the fragment header how is it possible for the code to come so far and how would it ever be possible to exceed fragments that are not there?

BourgondAries:
@TimonPost just try `cat file-with-those-bytes | nc -u -w1 localhost <port>` when setting up a laminar server. It crashes.

PR #255, opened 2019-09-25
Removed SendError that can never occur.

fraillt:
Simple as that.
[`SendError`](https://docs.rs/crossbeam-channel/0.1.2/crossbeam_channel/struct.SendError.html) can never happen within laminar, because of crossbeam's `Sender` and `Receiver` always exists together.
> ... can **only** fail if the receiving end of a channel is disconnected ...

Breaking changes in API:
* Changed "pub fn send(&mut self, packet: Packet)~~->Result<()>~~" in `Socket`.
* Removed `SendError` from `ErrorKind`.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -7,11 +7,10 @@ on:
     
 jobs:
   test:
-    name: Test on node ${{ matrix.node_version }} and ${{ matrix.os }}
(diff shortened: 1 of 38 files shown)


TimonPost:
Your correct here. There is one consern that I have. This is causing the the api to break. And maybe in the future we do have some error here and then we need to add the result back. Although I'm not sure if that is something we should care about now. @jstnlf?

PR #257, opened 2019-09-26
clippy warnings for tests

fraillt:
Fixed `cargo clippy --tests`.
Non-trivial fixes include:
* few floats compares. (error: `strict comparison of f32 or f64`)
* few macros were rewritten. (warning: `the function has a cognitive complexity of <N>`)

### DIFF
--- a/src/infrastructure/arranging/ordering.rs
+++ b/src/infrastructure/arranging/ordering.rs
@@ -432,22 +432,10 @@ mod tests {
         ( [$( $x:expr ),*] , [$( $y:expr),*] , $stream_id:expr) => {
         {
             // initialize vector of given range on the left.
-            let mut before: Vec<u16> = Vec::new();
-            $(
-                before.push($x);
-            )*
+            let before = [$($x,)*];
 
             // initialize vector of given range on the right.
-            let mut after: Vec<u16> = Vec::new();
-            $(
-                after.push($y);
-            )*
-
-            // generate test packets
-            let mut packets = Vec::new();
-            for (_, v) in before.iter().enumerate() {
-                packets.push(Packet::new(*v, $stream_id));
-            }
+            let after = [$($y,)*];
 
             // create system to handle the ordering of our packets.
             let mut ordering_system = OrderingSystem::<Packet>::new();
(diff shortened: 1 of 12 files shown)


src/net/quality.rs
```diff
@@ -119,6 +119,6 @@ mod test {
         let smoothed_rtt = network_quality.smooth_out_rtt(300);
 
         // 300ms has exceeded 50ms over the max allowed rtt. So we check if or smoothing factor is now 10% from 50.
-        assert_eq!(smoothed_rtt, 5.0);
+        assert_eq!((smoothed_rtt - 5.0f32).abs() < std::f32::EPSILON, true);
```
TimonPost: what is this doing?

fraillt: Clippy shows an error if I try to compare two floats for equality, and it should.
So what I do, is check if their absolute difference is smaller than epsilon.
It only returns true if smoothed_rtt is between ~4.99999... and ~5.00000001...

src/infrastructure/arranging/ordering.rs
```diff
@@ -432,49 +432,28 @@ mod tests {
...
-            $(
-                after.push($y);
-            )*
-
-            // generate test packets
-            let mut packets = Vec::new();
-            for (_, v) in before.iter().enumerate() {
-                packets.push(Packet::new(*v, $stream_id));
-            }
+            let after = [$($y,)*];
```
TimonPost: Maybe I'm missing something but I can't see the place were after is filled with packets?

src/infrastructure/arranging/sequencing.rs
```diff
@@ -185,23 +185,10 @@ mod tests {
...
-                $(
-                    after.push($y);
-                )*
-
-                // generate test packets
-                let mut packets = Vec::new();
-
-                for (_, v) in before.iter().enumerate() {
-                    packets.push(Packet::new(*v, $stream_id));
-                }
```
TimonPost: Same here

src/infrastructure/arranging/ordering.rs
fraillt: `after` is not filled with packets, but with `u16`s.
so instead of
```rust
let mut after: Vec<u16> = Vec::new();
$(
  after.push($y);
)*
```
I simply wrote `let after = [$($y,)*];`

If your asking why I skip this:
```rust
let mut packets = Vec::new();
for (_, v) in before.iter().enumerate() {
  packets.push(Packet::new(*v, $stream_id));
}
```
I decided to construct packet in place, so instead of `stream.arrange(packet.sequence, packet.clone())` i simply construct it directly `stream.arrange(*seq, Packet::new(*seq, $stream_id))`

src/infrastructure/arranging/sequencing.rs
fraillt: Same principles applied here, as in `src/infrastructure/arranging/ordering.rs`

src/infrastructure/arranging/ordering.rs
TimonPost: Oke thanks for the comments

PR #258, opened 2019-10-03
Codebase improvements.

fraillt:
This PR doesn't add or remove any existing functionality or tests, it only solves a few issues in a codebase.

1) Better interface separation with clear functionality boundaries.
* PROBLEM: There are no clear boundary between `Socket` and `ActiveConnections`.
`Socket` know too many details of a connection and `ActiveConnections` is basically a "proxy" object. Since a lot of functionality is in a `Socket`, it's impossible to write continuations, because of the Rust borrowing rules. E.g. to send heartbeat packets, you have to collect them first (heap allocation) and then send them, instead of iterating over packets and sending immediately (no heap allocation).
* SOLUTION: New types: `SocketController` and `ConnectionController` is introduced.
  * `SocketController` owns all core components: active connections, connection handler, event sender/receiver, socket sender/receiver. It is connection agnostic, it only stores a list of active connections, but doesn't do any operations on them. It only has one method `manual_poll` that does something useful: gets packets and events, and pass all connection related actions to a `ConnectionController` along with a connection.
  * `ConnectionController` Controls all aspects of the connection and owns only required components to process it: 
    * Processes incoming data (from a socket) or events (from a user).
    * Updates connection state: resends dropped packets, sends heartbeat packet, etc.
    * Creates new connections.
    * Checks if the connection should be dropped.

2) Separated unit tests vs integration tests. 
* PROBLEM: There are multiple issues with tests located in `src/net/socket.rs`: 
  * Sometimes some tests just randomly fail. They are like unit tests, while at the same time they are integration tests because they use real implementation of UdpSocket, that depends on a state of an OS.
  * Some tests are testing things, that should instead be tested elsewhere. Few examples: 
    * `fragmentation_send_returns_right_size` this is connection detail, and `Socket` shouldn't care about packet size or what is "fragmented" packet, etc... 
    * `ordered_16_bit_overflow` also a connection detail. More importantly, since it is computation heavy it takes ~11sec on my machine to run it, because of the overhead by testing through `Socket` interface instead of testing directly.
* SOLUTION: By convention, unit tests shouldn't depend on any external components, they either fail or succeed 100% of the time. Rust, [by convention](https://doc.rust-lang.org/book/ch11-03-test-organization.html), separates unit tests and integration tests in different directories. 
In order to make a majority of tests a unit tests, few traits were introduced: `SocketSender` and `SocketReceiver` and by using these traits `SocketController` and `ConnectionController` became socket type agnostic. So actual changes involved:
  * majority of tests was converted to unit tests by using fake socket implementation. 
  * some tests, that was actually integration tests was improved and moved to `test/basic_socket_test.rs`.
  * some tests were moved to `VirtualConnection` because they were more related to an actual connection.

BREAKING CHANGES:
* Removed `set_link_conditioner` from `Socket`. It doesn't make sense to provide this test function in public API. Why would any sane user need to drop some random packets?

Although this looks like a big PR, it actually reduced the number of relevant code. By moving different parts to different places, it allowed writing cleaner code.

This is the last PR to prepare for #246 integration.

**EDIT**
A lot of [changes](https://github.com/amethyst/laminar/pull/258#issuecomment-539511762) were made during the process


src/net/connection_controller.rs
```diff
@@ -0,0 +1,158 @@
...
+
+/// Defines a connection type.
+type Connection = VirtualConnection;
+/// Defines a user event type.
+type UserEvent = Packet;
+/// Defines a connection event type.
+type ConnectionEvent = SocketEvent;
+
+impl<PacketSender: SocketSender> ConnectionController<PacketSender> {
+    /// Creates a new instance of `ConnectionHandler`.
```
BourgondAries: Should be ConnectionController.

src/net/virtual_connection.rs
```diff
@@ -939,4 +938,115 @@ mod tests {
...
+
+            for (packet, _) in packets.into_iter() {
+                let mut recv_buff = [0; 4];
+                recv_buff.copy_from_slice(packet.payload());
+                let value = u32::from_ne_bytes(recv_buff);
+                assert_eq!(value, last_recv_value + 1);
+                last_recv_value = value;
+            }
+        }
+        assert_eq![last_recv_value, 99999];
```
BourgondAries: 99_999

src/lib.rs
```diff
@@ -37,7 +37,10 @@ mod throughput;
 #[cfg(feature = "tester")]
 pub use self::throughput::ThroughputMonitoring;
 
+#[cfg(test)]
+pub mod test_utils;
+
 pub use self::config::Config;
 pub use self::error::{ErrorKind, Result};
-pub use self::net::{LinkConditioner, Socket, SocketEvent};
```
TimonPost: When users want to simulate packet drop, they should be able to do so. Maybe not within the public API but behind a feature. Where is this `LinkConditioner` moved to?

src/net/connection_controller.rs
```diff
@@ -0,0 +1,158 @@
...
+};
+use crossbeam_channel::Sender;
+use log::error;
+use std::{self, net::SocketAddr, time::Instant};
+
+/// Controls all aspects of the connection:
+/// * Processes incoming data (from a socket) or events (from a user).
+/// * Updates connection state: resends dropped packets, sends heartbeat packet, etc.
+/// * Creates new connections.
+/// * Checks if connection should be dropped.
```
TimonPost:
```suggestion
/// * Checks if the connection should be dropped.
```

fraillt:
I'm not entirely happy with `SocketController` and `ConnectionController` names either :) During development I used `SocketHandler` and `ConnectionHandler`, but decided that controller was abetter fit.
It is the hardest part, to properly name these types and I need your help on this.

src/net/connection_controller.rs
```diff
@@ -0,0 +1,158 @@
...
+        event_sender: Sender<ConnectionEvent>,
+    ) -> Self {
+        ConnectionController {
+            config,
+            packet_sender,
+            event_sender,
+        }
+    }
+
+    /// Creates new connection. Also will init it and send connection event to a user.
```
TimonPost:
```suggestion
    /// Creates new connection and initialize it by sending an connection event to the user
```

src/net/connection_controller.rs
```diff
@@ -0,0 +1,158 @@
...
+        &self,
+        address: SocketAddr,
+        time: Instant,
+        initial_data: Option<&[u8]>,
+    ) -> Connection {
+        // emit connect event if this is initiated by remote host.
+        if initial_data.is_some() {
+            self.event_sender
+                .send(ConnectionEvent::Connect(address))
+                .unwrap();
```
TimonPost: I prefer except with an exception message

src/net/connection_controller.rs
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /// Determine if this connection should be dropped due to its state.
+    pub fn should_drop(&self, connection: &mut Connection, time: Instant) -> bool {
+        let should_drop = connection.packets_in_flight() > self.config.max_packets_in_flight
+            || connection.last_heard(time) >= self.config.idle_connection_timeout;
+        if should_drop {
+            self.event_sender
+                .send(ConnectionEvent::Timeout(connection.remote_address))
+                .unwrap();
```
TimonPost: I prefer except with an exception message

src/net/connection_controller.rs
```diff
@@ -0,0 +1,158 @@
...
+        }
+        should_drop
+    }
+
+    /// Handle a packet received from a socket.
```
TimonPost: maybe we test laminar over an in-memory transport layer, it isn't a socket always.

src/net/connection_controller.rs
```diff
@@ -0,0 +1,158 @@
...
+                    self.event_sender
+                        .send(ConnectionEvent::Packet(incoming.0))
+                        .unwrap();
+                }
+            }
+            Err(err) => error!("Error occured processing incomming packet: {:?}", err),
+        }
+    }
+
+    /// Handle an event received from a user.
```
TimonPost: Can we make more specific comments, I am not sure what we can understand under 'Handle' an event.

src/net/connection_controller.rs
```diff
@@ -0,0 +1,158 @@
...
+                self.send_packets(
+                    &connection.remote_address.clone(),
+                    connection.process_outgoing(PacketInfo::heartbeat_packet(&[]), None, time),
+                    "heatbeat packet",
+                );
+            }
+        }
+    }
+
+    /// Helper function that sends multiple outgoing packets
```
TimonPost:
```suggestion
    /// Sends multiple outgoing packets.
```

src/net/connection_controller.rs
TimonPost: Also, what do you mean with `Handle`

src/net/connection_controller.rs
```diff
@@ -0,0 +1,158 @@
...
+                    // This is stored with the dropped packet because they could be mixed
+                    ordering: dropped.ordering_guarantee,
+                },
+                dropped.item_identifier,
+                time,
+            );
+            self.send_packets(&connection.remote_address, packets, "dropped packets");
+        }
+
+        // send heartbeat packets if required.
```
TimonPost:
```suggestion
        // send heartbeat packets if required
```

src/net/connection_controller.rs
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /// Creates new connection. Also will init it and send connection event to a user.
+    pub fn create_connection(
+        &self,
+        address: SocketAddr,
+        time: Instant,
+        initial_data: Option<&[u8]>,
+    ) -> Connection {
+        // emit connect event if this is initiated by remote host.
```
TimonPost:
```suggestion
        // Emit connect event if this is initiated by the remote host
```

src/net/connection_controller.rs
TimonPost: And describe what the parameters mean, like initial data, time.

src/net/connection_controller.rs
```diff
@@ -0,0 +1,158 @@
...
+        if initial_data.is_some() {
+            self.event_sender
+                .send(ConnectionEvent::Connect(address))
+                .unwrap();
+        }
+        Connection::new(address, &self.config, time)
+    }
+
+    /// Determine if this connection should be dropped due to its state.
```
TimonPost:
```suggestion
    /// Determine if the given `Connection` should be dropped due to its state.
```

src/net/connection_controller.rs
```diff
@@ -0,0 +1,158 @@
...
+            "user packet",
+        );
+    }
+
+    /// Process various connection related tasks: resend dropped packets, send heartbeat packet, etc...
+    /// This function gets called very frequently.
+    pub fn update(&mut self, connection: &mut Connection, time: Instant) {
+        // resend dropped packets
+        let dropped_packets = connection.gather_dropped_packets();
+        for dropped in dropped_packets {
```
TimonPost: combine into: 
```
for dropped in connection.gather_dropped_packets() {
```

src/test_utils/fake_socket.rs
```diff
@@ -0,0 +1,69 @@
...
+        self.handler.manual_poll(time);
+    }
+
+    /// Returns a number of active connections.
+    pub fn connection_count(&self) -> usize {
+        self.handler.connections_count()
+    }
+
+    /// Set the link conditioner for this socket. See [LinkConditioner] for further details.
+    pub fn set_link_conditioner(&mut self, conditioner: Option<LinkConditioner>) {
```
TimonPost: We don't export `LinkConditioner` right, or do we?

src/test_utils/fake_socket.rs
```diff
@@ -0,0 +1,69 @@
...
+    pub fn manual_poll(&mut self, time: Instant) {
+        self.handler.manual_poll(time);
+    }
+
+    /// Returns a number of active connections.
+    pub fn connection_count(&self) -> usize {
+        self.handler.connections_count()
+    }
+
+    /// Set the link conditioner for this socket. See [LinkConditioner] for further details.
```
TimonPost:
```suggestion
    /// Sets the link conditioner for this socket. See [LinkConditioner] for further details.
```

src/test_utils/fake_socket.rs
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /// Send a packet
+    pub fn send(&mut self, packet: Packet) -> Result<()> {
+        // we can savely unwrap, because receiver will always exist
+        self.handler.event_sender().send(packet).unwrap();
+        Ok(())
+    }
+
+    /// Receive a packet
```
TimonPost:
```suggestion
    /// Receives a packet.
```

src/test_utils/fake_socket.rs
```diff
@@ -0,0 +1,69 @@
...
+    pub fn recv(&mut self) -> Option<SocketEvent> {
+        if let Ok(event) = self.handler.event_receiver().try_recv() {
+            Some(event)
+        } else {
+            None
+        }
+    }
+
+    /// Process any inbound/outbound packets and handle idle clients
```
TimonPost:
```suggestion
    /// Process any inbound/outbound packets and handle idle clients.
```

src/test_utils/network_emulator.rs
```diff
@@ -0,0 +1,113 @@
...
+type GlobalBindings = Rc<RefCell<HashMap<SocketAddr, VecDeque<(SocketAddr, Vec<u8>)>>>>;
+
+/// Enables to create the emulated socket, that share global state stored by this network emulator.
+#[derive(Debug, Default)]
+pub struct NetworkEmulator {
+    network: GlobalBindings,
+}
+
+impl NetworkEmulator {
+    /// Created an emulated socket by binding to a address.
```
TimonPost:
```suggestion
    /// Creates an emulated socket by binding to an address.
```

src/test_utils/network_emulator.rs
```diff
@@ -0,0 +1,113 @@
...
+/// Implementation of a socket, that is created by `NetworkEmulator`.
+#[derive(Debug, Clone)]
+pub struct EmulatedSocket {
+    network: GlobalBindings,
+    address: SocketAddr,
+    conditioner: Option<LinkConditioner>,
+}
+
+impl EmulatedSocket {
+    /// Set the link conditioner for this socket. See [LinkConditioner] for further details.
```
TimonPost:
```suggestion
    /// Sets the link conditioner for this socket. See [LinkConditioner] for further details.
```

src/test_utils/network_emulator.rs
```diff
@@ -0,0 +1,113 @@
...
+            {
+                let slice = &mut buffer[..payload.len()];
+                slice.copy_from_slice(payload.as_ref());
+                Some((slice, addr))
+            } else {
+                None
+            },
+        )
+    }
+    /// Returns the socket address that this socket was created from.
```
TimonPost: new line above

src/net/socket_controller.rs
```diff
@@ -0,0 +1,850 @@
...
+}
+
+/// This trait can be implemented to receive data from the socket.
+pub trait SocketReceiver: Debug {
+    /// Receives a single packet from the socket.
+    fn receive_packet<'a>(
+        &mut self,
+        buffer: &'a mut [u8],
+    ) -> Result<Option<(&'a [u8], SocketAddr)>>;
+    /// Returns the socket address that this socket was created from.
```
TimonPost: new line above

src/net/socket_controller.rs
```diff
@@ -0,0 +1,850 @@
...
+    fn local_addr(&self) -> Result<SocketAddr>;
+}
+
+/// A reliable generic socket implementation with configurable reliability and ordering guarantees.
+/// Owns all core components: active connections, connection handler, event sender/receiver, socket sender/receiver.
```
TimonPost: Is this a public exported type, so yes we probably don't have to mention what's inside this.

src/net/socket_controller.rs
```diff
@@ -0,0 +1,850 @@
...
+                }
+                Ok(None) => break,
+                Err(e) => error!("Encountered an error receiving data: {:?}", e),
+            }
+            if self.is_blocking_mode {
+                break;
+            }
+        }
+
+        // Now grab all the packets waiting to be sent and send them
```
TimonPost:
```suggestion
        // Now grab all the waiting packets and send them
```

src/net/socket_controller.rs
```diff
@@ -0,0 +1,850 @@
...
+                .or_insert_with(|| handler.create_connection(event.addr(), time, None));
+            handler.handle_event(conn, event, time);
+        }
+
+        // Update all connections
+        for conn in self.connections.values_mut() {
+            handler.update(conn, time);
+        }
+
+        // Iterate through all connections and remove those that should be dropped.
```
TimonPost:
```suggestion
        // Iterate through all connections and remove those that should be dropped
```

src/net/socket_controller.rs
```diff
@@ -0,0 +1,850 @@
...
+        self.connections
+            .retain(|_, conn| !handler.should_drop(conn, time));
+    }
+
+    pub fn local_addr(&self) -> Result<SocketAddr> {
```
TimonPost: doc string

fraillt:
`Socket` is not renamed, it basically wraps `SocketController`.

src/lib.rs
fraillt: `LinkConditioner` was moved to `test_utils/link_conditioner.rs`.
I like the idea to hide it behind the feature flag, it makes more sense.

src/net/connection_controller.rs
fraillt: By `handle` I ment something that processes, maybe my brain hasn't switched from another project, where I see lots of "handle_request" and stuff. I rename it to `process` :)

src/net/connection_controller.rs
fraillt: This is not a public interface, I'm not sure if it requires a lot of details... and there is not many of them
Would "Process a received event and send a packet." be better?

src/net/connection_controller.rs
fraillt: Is this comment ok? 
`Process a received packet: parse it and emit an event.`?

src/net/socket_controller.rs
fraillt: Ok, you're right it doesn't convey any useful information.

src/test_utils/fake_socket.rs
fraillt: No, it is only used in tests.

src/net/connection_controller.rs
TimonPost: Better, it is just an comment to be specific in comments. By saying parsing and emitting an event, it makes much more sense.

fraillt:
Build is failing in jenkins, probably due to older clippy version.
Maybe could add addition jenkins stage, "Print Environment" before "Check Formatting"?
that would have few commands like:
```bash
cargo --version
cargo clippy --version
cargo fmt --version
```
What do you think?

TimonPost:
Yea we should have something like that

fraillt:
I tried to add new stage "Environment Info" in Jenkinsfile, but it didn't appeared in CI.
So I restored it.

src/net/connection_controller.rs
```diff
@@ -0,0 +1,162 @@
...
+/// It doesn't own connections, but only owns necessary components to process them.
+#[derive(Debug)]
+pub struct ConnectionController<PacketSender> {
+    config: Config,
+    packet_sender: PacketSender,
+    event_sender: Sender<SocketEvent>,
+}
+
+/// Defines a connection type.
+type Connection = VirtualConnection;
```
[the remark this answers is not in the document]
fraillt: I thought that I will introduce these types in the connection manager PR, so I wouldn't need to change signatures of the functions, just remove the type aliases :)

src/test_utils/fake_socket.rs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /// Returns a handle to the event receiver which provides a thread-safe way to retrieve events
+    /// from the socket. This should be used when the socket is busy running its polling loop in
+    /// a separate thread.
+    pub fn get_event_receiver(&self) -> Receiver<SocketEvent> {
+        self.handler.event_receiver().clone()
+    }
+
+    /// Send a packet.
```
TimonPost:
```suggestion
    /// Sends a packet.
```

src/net/socket_controller.rs
```diff
@@ -0,0 +1,868 @@
...
+    fn send_packet(&mut self, addr: &SocketAddr, payload: &[u8]) -> Result<usize>;
+}
+
+/// This trait can be implemented to receive data from the socket.
+pub trait SocketReceiver: Debug {
+    /// Receives a single packet from the socket.
+    fn receive_packet<'a>(
+        &mut self,
+        buffer: &'a mut [u8],
+    ) -> Result<Option<(&'a [u8], SocketAddr)>>;
```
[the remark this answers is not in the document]
fraillt: OK, will do that

src/test_utils/fake_socket.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub fn send(&mut self, packet: Packet) -> Result<()> {
+        // we can savely unwrap, because receiver will always exist
+        self.handler.event_sender().send(packet).unwrap();
+        Ok(())
+    }
+
+    /// Receive a packet.
```
TimonPost:
```suggestion
    /// Receives a packet.
```

src/test_utils/fake_socket.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub fn recv(&mut self) -> Option<SocketEvent> {
+        if let Ok(event) = self.handler.event_receiver().try_recv() {
+            Some(event)
+        } else {
+            None
+        }
+    }
+
+    /// Process any inbound/outbound packets and handle idle clients.
```
TimonPost:
```suggestion
    /// Processes any inbound/outbound packets and handle idle clients.
```

src/test_utils/network_emulator.rs
```diff
@@ -0,0 +1,112 @@
...
+            }
+            Ok(payload.len())
+        } else {
+            Ok(0)
+        }
+    }
+}
+
+impl SocketReceiver for EmulatedSocket {
+    /// Receive a packet from this socket.
```
TimonPost:
```suggestion
    /// Receives a packet from this socket.
```

src/net/socket.rs
```diff
@@ -1,87 +1,94 @@
...
-    receiver: Receiver<SocketEvent>,
-    sender: Sender<Packet>,
-}
-
-enum UdpSocketState {
-    MaybeEmpty,
-    MaybeMore,
+    // Stores an instance of `SocketController` where `SocketSender` uses a `UdpSocket` (with `LinkConditioner`, if enabled) and SocketReceiver` is a `UdpSocket`.
+    handler: SocketController<SocketWithConditioner, UdpSocket>,
+    link_conditioner: Arc<Mutex<Option<LinkConditioner>>>,
```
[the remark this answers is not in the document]
fraillt: I tried to delegate it, but it looks "incorrect" 
I need to pass delegate it through a SocketImpl as well as ConnectionController, and they do not care at all about this link_conditioner. I see that we have two (three?) options here:
1) delegate through all interfaces, regardless if they care about it or not.
PRO: no performance penalty when a feature is enabled.
CON: the need to change interfaces, that doesn't care about this.
2) use `Arc<Mutex<....>>`
PRO: cleaner interfaces
CON: performance penalty when a feature is enabled.
3) Since we know that `LinkConditioner` will never be set from a different thread, maybe we can express that in rust, but I need your help here, I'm not very proficient in rust yet...
PRO: no (or minimum) performance penalty when a feature is enabled.
CON: ... maybe it might require to write "unsafe" code.

What should I do?

src/net/socket_controller.rs
```diff
@@ -0,0 +1,868 @@
...
+                    } else {
+                        // create connection but do not add to active connections list
+                        let mut conn = handler.create_connection(address, time, Some(payload));
+                        handler.process_packet(&mut conn, payload, time);
+                    }
+                }
+                Ok(None) => break,
+                Err(e) => error!("Encountered an error receiving data: {:?}", e),
+            }
+            if self.is_blocking_mode {
```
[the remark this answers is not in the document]
fraillt: There is no change actually.
If a socket is in blocking mode, then return after a first iteration.

src/net/connection_controller.rs
```diff
@@ -0,0 +1,162 @@
...
+/// Defines a user event type.
+type UserEvent = Packet;
+/// Defines a connection event type.
+type ConnectionEvent = SocketEvent;
+
+impl<PacketSender: SocketSender> ConnectionController<PacketSender> {
```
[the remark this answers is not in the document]
fraillt: Good observation!
It does look confusing, it should be the other way around. The problem is that SocketController or SocketImpl is probably a bad name to start with...
`SocketSender` and `SocketReceiver` are defined together with `SocketImpl`, these names make sense in this context, and it looks like this:
`pub struct SocketImpl<TSender: SocketSender, TReceiver: SocketReceiver>`
I just got an idea, of how I change things around, to make it better. 
I'll write down more details when I do it.

fraillt:
Indeed, lower case vs upper case in function bodies are very similar.
I ran this in a root directory on master branch: 
`grep -r -i --include \*.rs " // "  | cut -d $' ' -f 2- | cut -d '/' -f 2- | cut -c 3-3 | sort | uniq.exe -c | sort`
* 104 starts with upper case.
* 87 starts with lower case.
Most of the comments are inside function bodies.

TimonPost:
Don't mind them for now. That can be done in another PR.

fraillt:
This commit addresses a lot of issues.
A lot of changes were made, but there is the gist of them are:
* `SocketImpl` is now `ConnectionManager<TSocket: DatagramSocket, TConnection: Connection>`. It is a fully generic type, that depends on a socket and connection.
* new trait `Connection`, it allows to implement a concept of connection, that is used by `ConnectionManager`, it also defines types of send and receive events.
* instead of separate `SocketSender` and `SocketReceiver` traits, one `DatagramSocket` was introduced, this allowed to simplify some things.

I feel that the current state is much better now, but there are a few things that I'm not entirely happy with:
* to preserve backward compatibility addition trait was introduced `ConnectionEventAddress`. This was a necessary step because `ConnectionManager` doesn't know anything about types of connection events. Code could be simplified further if all events will be wrapped in this type:
  ```rust
  pub struct ConnectionEvent<Event: Debug>(pub SocketAddr, pub Event)
  ```
* ALL methods of `Connection` accepts a special trait, that does 3 things:
  * returns a config
  * sends an event
  * sends a packet
  I named this trait a `ConnectionMessenger`... I could probably remove this entirely, but then all methods of a connection would accept additional 2 parameters for each method.

PR #260, opened 2019-10-11
Syncing documentation

TimonPost:
- [X] Synced '.' behind all sentences
- [X] Synced plural for functions 'creates', 'returns', etc.
- [x] Sync non-capital letter doc strings in code `//`

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,8 @@ and provides a lightweight, message-based interface which provides certain guara
 
 Laminar was designed to be used within the [Amethyst][amethyst] game engine but is usable without it.
 
+If you are new to laminar or networking in general, We strongly recommend taking a look at the [laminar book][book]
+
 [amethyst]: https://github.com/amethyst/amethyst
 
 # Concepts
@@ -35,7 +37,6 @@ For more information, read the projects [README.md][readme], [book][book], [docs
 [book]: https://amethyst.github.io/laminar/docs/index.html
 [docs]: https://docs.rs/laminar/
 [examples]: https://github.com/amethyst/laminar/tree/master/examples
-
 [amethyst]: https://github.com/amethyst/amethyst
 
 ## Table of contents:
@@ -61,6 +62,9 @@ These are the features this crate provides:
 - RTT Estimation
 - Link conditioner to simulate packet loss and latency
 - Well-tested by integration and unit tests
+- Basic DoS mitigation
+- High Timing control
+- Can be used by multiple threads (Sender, Receiver)
 
 ## Getting Stated
 Add the laminar package to your `Cargo.toml` file.
--- a/README.md
+++ b/README.md
@@ -52,26 +52,36 @@ For more information, read the projects [README.md][readme], [book][book], [docs
 ## Features
 These are the features this crate provides:
 
-- UDP-based Protocol
-- Connection Tracking
-- Automatic Fragmentation
-- Reliability Options: Unreliable and Reliable
-- Arranging Options: Sequenced, Unordered, and Ordered.
-- Arranging Streams
-- Protocol Versioning
-- RTT Estimation
-- Link conditioner to simulate packet loss and latency
-- Well-tested by integration and unit tests
-- Basic DoS mitigation
-- High Timing control
-- Can be used by multiple threads (Sender, Receiver)
+* [x] Fragmentation
+* [x] Unreliable packets
+* [x] Unreliable sequenced packets
+* [x] Reliable unordered packets
+* [x] Reliable ordered packets
+* [x] Reliable sequenced packets
+* [x] Fragmentation
+* [x] Rtt estimations
+* [x] Protocol version monitoring
+* [x] Basic connection management
+* [x] Heartbeat
+* [x] Basic DoS mitigation
+* [x] High Timing control
+* [x] Protocol Versioning
+* [x] Well-tested by integration and unit tests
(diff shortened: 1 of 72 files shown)


README.md
```diff
@@ -15,8 +15,8 @@
 [s6]: https://tokei.rs/b1/github/amethyst/laminar?category=code
 [s7]: https://codecov.io/gh/amethyst/laminar/branch/master/graphs/badge.svg
 
-Laminar is a semi-reliable UDP-based protocol for multiplayer games. This library implements wrappers around the UDP-protocol,
-and provides a lightweight, message-based interface which provides certain guarantees like reliability and ordering.
+Laminar is an application-level transport protocol that provides configurable reliability and ordering guarantees build on top of UDP.
```
jstnlef: Laminar is an application-level transport protocol which provides configurable reliability and ordering guarantees built on top of UDP.

README.md
```diff
@@ -15,8 +15,8 @@
 [s6]: https://tokei.rs/b1/github/amethyst/laminar?category=code
 [s7]: https://codecov.io/gh/amethyst/laminar/branch/master/graphs/badge.svg
 
-Laminar is a semi-reliable UDP-based protocol for multiplayer games. This library implements wrappers around the UDP-protocol,
-and provides a lightweight, message-based interface which provides certain guarantees like reliability and ordering.
+Laminar is an application-level transport protocol that provides configurable reliability and ordering guarantees build on top of UDP. 
+It is focused on fast-paced fps-games and provides a lightweight, message-based interface.
```
jstnlef: focuses on

docs/CONTRIBUTING.md
```diff
@@ -34,10 +34,46 @@ Some code guidelines to keep in mind when contributing to laminar or amethyst-ne
...
 2. Hard Coding
     - Don't hard code values anywhere
     - Use the ‘NetworkConfig’ type for common network settings, use consts or parameter input
     - Use of lazy_static is acceptable but first make sure you can’t fix the issue in other ways
 3. Code markup
     - Keep files small. Better have small files with small pieces of logic than having one file with 1000 lines of logic with multiple types/structs etc. Note that I speak of logic, tests not included
 	- No panics/unwraps in the main codebase, but they are accepted in tests
+
+## Import Reordering
+All imports are semantically grouped and ordered. The order is:
```
jstnlef: Do we have a way to enforce this? It seems awfully pedantic if we don't have a programmatic way of both performing the transform and validating it.

src/infrastructure/acknowledgment.rs
```diff
@@ -45,14 +45,14 @@ impl AcknowledgmentHandler {
...
-    /// Returns the ack_bitfield corresponding to which of the past 32 packets we've
+    /// Returns the `ack_bitfield` corresponding to which of the past 32 packets we've
     /// successfully received.
     pub fn ack_bitfield(&self) -> u32 {
         let most_recent_remote_seq_num: u16 = self.remote_sequence_num();
         let mut ack_bitfield: u32 = 0;
         let mut mask: u32 = 1;
 
-        // Iterate the past REDUNDANT_PACKET_ACKS_SIZE received packets and set the corresponding
+        // iterate the past `REDUNDANT_PACKET_ACKS_SIZE` received packets and set the corresponding
```
jstnlef: Curious why would wouldn't capitalize these as well?

src/lib.rs
```diff
@@ -1,5 +1,5 @@
-//! Laminar is a semi-reliable UDP-based protocol for multiplayer games. This library implements wrappers around the UDP-protocol,
-//! and provides a lightweight, message-based interface which provides certain guarantees like reliability and ordering.
+//! Laminar is an application-level transport protocol that provides configurable reliability and ordering guarantees build on top of UDP.
```
jstnlef: Same comment as above.

src/net/socket.rs
```diff
@@ -152,26 +152,26 @@ impl Socket {
         }
     }
 
-    /// Entry point to the run loop. This should run in a spawned thread since calls to `poll.poll`
-    /// are blocking. We will default this to sleeping for 1ms.
+    /// Runs the polling loop with the default '1ms' sleep duration. This should run in a spawned thread
+    /// since calls to `poll.poll` are blocking.
```
jstnlef: `self.manual_poll`

src/packet/process_result.rs
```diff
@@ -3,7 +3,7 @@ use crate::packet::{OutgoingPacket, Packet, PacketType};
 
 use std::collections::VecDeque;
 
-/// Struct that implements `Iterator`, and is used to return incoming (from bytes to packets) or outgoing (from packet to bytes) packets.
+/// Is used to return incoming (from bytes to packets) or outgoing (from packet to bytes) packets.
```
jstnlef: Used to return incoming

docs/CONTRIBUTING.md
TimonPost: I don't believe so. However, as noted the IntelIJ rust plugin is able to do this. So once in a while, we can run that to validate everything is ordered correctly. I have no idea how to enforce it in some way.

src/infrastructure/acknowledgment.rs
TimonPost: Because there are also very small '//' comments, and those are often not a complete sentence. I do think that either way we should do everything capitalized or with a lowercase. This PR has everything lower case. It is a matter of preference.

PR #261, opened 2019-10-12
Update commit email

jstnlef:


TimonPost:
What the heck is wrong with the git history. This PR and my pr's all have millions of commits. Going to do a rebase and see what the problem is here.

BourgondAries:
@TimonPost because it's a commit rewrite all other active commits will be having issues. This was to rewrite an email that shouldn't have been there.

PR #278, opened 2020-08-27
Fix #277 Add connections on inbound packets

### LINKED ISSUE #277: Connection events are fired before the connection is added
Heartbeat messages do not begin sending if a Packet is received, they will only begin sending after an outbound Packet is sent. This causes two issues:

1) The opposite side of the connection will perceive this as a Timeout event and terminate the connection (since it doesn't receive any heartbeat messages).

2) This side of the connection will see each received heartbeat as a new connection event, which leads to repeated Connection events.

An example of this can be found here: https://github.com/ncallaway/laminar-heartbeat-demo

This creates a server (:12350) and a client (:12355). Each are configured to send heartbeats every 2 seconds and timeout after 5 seconds.

```
idle_connection_timeout: Duration::from_secs(5),
heartbeat_interval: Some(Duration::from_secs(2)),
```

The client, on startup will send a `Ping` to the server. The server can optionally respond with a `pong` message.

What we see is that if the server responds with the `pong` message, heartbeats flow in both direction and the connection is maintained on both ends.

If the server does not respond with a pong message, heartbeats only flow from the client to the server, until the client times out. The server sends a new `Connection` event for each heartbeat that it receives, and never reports a timeout event after the client stops sending heartbeat messages.

### Wireshark Capture

ping-capture

### Server

```
$ cargo run -- -s
[164.4µs] Bound to 127.0.0.1:12350
        [2.43s] Connection from: 127.0.0.1:12355
[2.43s] Received "Ping" from V4(127.0.0.1)
        [4.42s] Connection from: 127.0.0.1:12355
        [6.42s] Connection from: 127.0.0.1:12355
```

### Client

```
$ cargo run -- -c
[147.0µs] Bound to 127.0.0.1:12355
        [269.30µs] Sent: Ping
        [5.00s] Timeout from: 127.0.0.1:12350
```

### Wireshark Capture

ping-pong-capture-1

### Server

```
$ cargo run -- -s -p
[116.8µs] Bound to 127.0.0.1:12350
        [1.32s] Connection from: 127.0.0.1:12355
[1.32s] Received "Ping" from V4(127.0.0.1)
        [1.32s] Sent: pong
        [64.33s] Timeout from: 127.0.0.1:12355
```

### Client

```
$ cargo run -- -c
[154.3µs] Bound to 127.0.0.1:12355
        [273.42µs] Sent: Ping
[24.87ms] Received "pong" from V4(127.0.0.1)

# client interrupted
```

ncallaway:
This is a proposed fix for #277.

I've confirmed that the behavior of the [heartbeat demo](https://github.com/ncallaway/laminar-heartbeat-demo) seems more stable after this change:

**server**
```
$ cargo run -- -s                                                                                                 
[171.1µs] Bound to 127.0.0.1:12350
        [9.39ms] Connection from: 127.0.0.1:12355
[17.26ms] Received "Ping" from V4(127.0.0.1)
```

**client**
```
$ cargo run -- -c
[153.3µs] Bound to 127.0.0.1:12355
        [272.50µs] Sent: Ping
        [5.00s] Timeout from: 127.0.0.1:12350
```

I'm somewhat wary of this change due to the comment from the original refactor (https://github.com/amethyst/laminar/pull/258). I'd love to have @TimonPost or @fraillt give critical thought and feedback about whether this approach might cause any issues.

### DIFF
--- a/src/net/connection_manager.rs
+++ b/src/net/connection_manager.rs
@@ -111,6 +111,7 @@ impl<TSocket: DatagramSocket, TConnection: Connection> ConnectionManager<TSocket
                         let mut conn =
                             TConnection::create_connection(messenger, address, time, Some(payload));
                         conn.process_packet(messenger, payload, time);
+                        self.connections.insert(address, conn);
                     }
                 }
                 Err(e) => {


fraillt:
As I mentioned in #277 the main idea was to accept only by responding from the server. I think it is a security thing, as it allows server to select connections that it wants to accept, instead of someone else connecting without permission.

ncallaway:
I'm closing this based on @fraillt's comments and the discussion in #277.

The longer term resolution for this will be implementing a proper handshake. The shorter term resolution will likely be documentation updates.

PR #280, opened 2020-08-30
Resolve Issue #277: Change Connection behavior to require both a packet send and receive

ncallaway:
This PR is based on the discussion in https://github.com/amethyst/laminar/issues/277.

This change:

- Canonicalizes an "established connection" as requiring both a send and a receive from a given endpoint.
- Changes `SocketEvent::Connect` to only fire when a connection has been established.
- Changes heartbeats to only send after a connection has been established.
- Stops creating new `VirtualConnection`s from inbound messages, if we have too many unestablished connections already being tracked. `max_unestablished_connections`.
- Adds a `max_unestablished_connections` config option.
- Adds `SocketEvent::Disconnect` to fire when a `Timeout` fires from an endpoint with an established connection.
- Updates the unit tests to match the new established connection behavior.
- Updates the documentation around the `SocketEvent`s, and adds a `Connection` section to the mdbook.

**Open Questions**
* I'm not convinced about the `Disconnect` SocketEvent. It seemed useful to be to both mirror the `Connect` event, and to disambiguate a `Timeout` from an established connection being dropped. Still, it's one more `SocketEvent` that users need to be aware of and manage. As an alternative, we could modify `Timeout` to include a flag to indicate whether it was an established connection that timed out. As a third option, we could just drop this event all together and require users to know which endpoints are established connections and which aren't.
* I'm not sure if `Connections` should have their own section in the book. I decided to make it it's own section, because it seems like this is a current area of confusion for Laminar users. I'm happy to make changes or document this somewhere else.


docs/md_book/src/protocols.md
```diff
@@ -1,36 +1,41 @@
 # Networking protocols
 
-So first and possibly the important one is which protocol to use and when. Let’s first take a look at TCP and UDP.
+When building any networked application the first and possibly the important decision to make is which protocol to use and when. Laminar is built on top of UDP. Let’s first take a quick look at both TCP and UDP, then we'll explain why Laminar uses UDP.
```
ncallaway, line 3: This is the only semantic change I made to this file. The main reason for the change was I wanted to explicitly state that Laminar is built using UDP, because I wanted to mention that in the "Connections" section.

Everything below this is my markdown autoformatting changes. I'm happy to revert those changes below if they're undesirable.

src/net/connection_impl.rs
```diff
@@ -44,15 +45,15 @@ impl Connection for VirtualConnection {
...
-        initial_data: Option<&[u8]>,
     ) -> VirtualConnection {
-        // emit connect event if this is initiated by the remote host.
-        if initial_data.is_some() {
-            messenger.send_event(&address, SocketEvent::Connect(address));
-        }
         VirtualConnection::new(address, messenger.config(), time)
     }
 
+    /// Connections are considered established once they have both had both a send and a receive.
```
TimonPost, line 52:
```suggestion
    /// Connections are considered established once they both have had a send and a receive.

```

src/net/events.rs
```diff
@@ -7,10 +7,17 @@ use crate::packet::Packet;
 pub enum SocketEvent {
     /// A packet was received from a client.
     Packet(Packet),
-    /// A new client connected.
-    /// Clients are uniquely identified by the ip:port combination at this layer.
+    /// A new connection has been established with a client. A connection is considered
+    /// established whenever a packet has been both _sent_ and _received_ from the client.
+    ///
+    /// On the server—in order to receive a `Connect` event—you must respond to the the first
```
TimonPost, line 13:
```suggestion
    /// On the server—in order to receive a `Connect` event—you must respond to the first
```

src/net/virtual_connection.rs
```diff
@@ -54,6 +59,28 @@ impl VirtualConnection {
         }
     }
 
+    /// Record that this connection has sent a packet. Returns whether the connection has
```
TimonPost, line 62:
```suggestion
    /// Records that this connection has sent a packet. Returns whether the connection has
```

src/net/virtual_connection.rs
```diff
@@ -54,6 +59,28 @@ impl VirtualConnection {
...
+    /// become acknowledged because of this send.
+    pub fn record_send(&mut self) -> bool {
+        let was_est = self.is_established();
+        self.ever_sent = true;
+
+        !was_est && self.is_established()
+    }
+
+    /// Record that this connection has sent a packet. Returns whether the connection has
```
TimonPost, line 71:
```suggestion
    /// Records that this connection has sent a packet. Returns whether the connection has
```

Plural start for function comment is a code style in lamianr.

src/net/virtual_connection.rs
TimonPost: Plural start for function comment is a code style in lamianr.

src/net/connection_impl.rs
```diff
@@ -66,6 +67,12 @@ impl Connection for VirtualConnection {
                 &self.remote_address,
                 SocketEvent::Timeout(self.remote_address),
             );
+            if self.is_established() {
+                messenger.send_event(
+                    &self.remote_address,
+                    SocketEvent::Disconnect(self.remote_address),
```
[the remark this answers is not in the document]
ncallaway: The reason I left Timeout (or rather create `Disconnect`), was that I thought it was useful to have `Timeout` as a concept even if a connection is never established.

For example, as a user if I send a `Hello` to a server, and I never receive a reply, I need some way to be notified of that by Laminar. If we dropped `Timeout` entirely from this PR, then there would be no way of seeing from `Laminar` that we had no response.

So the semantics in this PR are that `Timeout` is _any_ timeout, whether connected or not, and `Disconnect` is a timeout where we previously had a connection.

I'm also a little uncomfortable with the semantics here, and I'm happy to change things, but if we drop `Timeout`, I think we need to have some way of notifying the user that we sent a `Hello`, and never heard back.

src/net/virtual_connection.rs
ncallaway: Good to know!

ncallaway:
I've incorporated all the documentation updates and dead-code removal. I'm also totally happy if we want to work through the `Timeout` / `Disconnect` issue before merging.

ncallaway:
Does anyone know why the `pr-merge` has been running indefinitely? I'm not really clear what that is testing, but it seems to have hung in some way.

TimonPost:
@ncallaway It seems like the windows bare metal machine is down. It might be the case that amethyst has switched CI's I am not sure.

TimonPost:
So regarding the extra socket event. I am fine with it now, after reading your reasons. It seems like a useful thing to provide. If the user doesn't want to handle it it can just use `_ => ` in its match. So that isn't a big issue. Can the PR be merged?

ncallaway:
@TimonPost I'm good with this being merged if the CI issue isn't a blocking problem. I'm also happy to wait until we resolve whatever the CI issue is, but I'm not sure I can be much assistance there.

TimonPost:
I;lll test it on windows then we can do a merge.

TimonPost:
Sorry for the delay, my week was filled with lots of other things. I tested it on windows, and it seems to behave good! So lets merge.

PR #286, opened 2021-02-28
Expose ConnectionManager and VirtualConnection

smokku:
This change exposes `ConnectionManager` and `VirtualConnection` (and their dependencies) for public consumption.
It allows for integrating Laminar's connection handling and different Packet delivery guarantees with custom socket implementation.

For an example of running Laminar over [naia-socket](https://github.com/naia-rs/naia-socket) (webrtc-unreliable) see [soldank/server/networking.rs](https://github.com/smokku/soldank/blob/master/server/src/networking.rs#L26) / [soldank/client/networking.rs](https://github.com/smokku/soldank/blob/master/client/src/networking.rs#L42)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,7 +24,10 @@
 
 pub use self::config::Config;
 pub use self::error::{ErrorKind, Result};
-pub use self::net::{LinkConditioner, Socket, SocketEvent};
+pub use self::net::{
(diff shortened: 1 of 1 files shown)


erlend-sh:
Afaik no one is actively maintaining Laminar right now, so if you don’t get feedback from anyone in the next 48hrs just ping me and I’ll give you commit rights.

TimonPost (approved):
Looks good to me. I am not actively maintaining laminar anymore, but I do not botter exposing this socket abstraction API.

Issue #290, opened 2021-05-31
Proto-RFC: Futures support in Laminar


james7132:
## Motivation
Provide a reliable UDP implementation that works well in async executors for games (i.e. `bevy_tasks`).

## Design
The cornerstone of this design calls for the heavy use of a [bidirectional version](https://github.com/HouraiTeahouse/backroll-rs/blob/main/backroll_transport/src/channel.rs#L4) of `async-channel` to pass messages around independent tasks on a futures executor. This acts as three things: a representation of the connection state (is the channel still open or not), a buffered connection to pass messages along, and an point of abstraction. Channels are generic only on the type of message they pass, so there is no need to create nested structs with generic parameters to connect up two distinct layers of the network stack (a la `ConnectionManager`).

A specialization of this channel, known in backroll_transport as a `Peer`, is a newtype around BidirectionalAsyncChannel that solely focuses on `Box<[u8]>`, raw binary packets. By default, Peers are assumed to be unreliable, unordered packets: a direct UDP analogue. This has already been implemented into backroll_transport, and is used as the main abstraction point for the actual I/O layer (see [backroll_transport_udp](https://docs.rs/backroll_transport_udp/0.1.0/backroll_transport_udp/struct.UdpManager.html)).

To help manage multiple ongoing connections, backroll_transport has a `Peers<T>`, a newtype wrapper around `DashMap<T, Peer>`, which only presents active and open connections by removing non-connected peers from public access. Dropping the entire `Peers<T>` object also closes all associated `Peer`s.

By using infinitely looping tasks around these bidrectional channels, it's possible to create a graph of interconnected tasks that construct the connection itself. Below is an example of one such connection (each box is a independent task, all outgoing and incoming arrows are separate multiple producer, multiple consumer channel):

![image](https://user-images.githubusercontent.com/3137680/120247119-4d389900-c227-11eb-9d19-805d53ef5765.png)

[...]

PR #291, opened 2021-06-03
Async I/O based Laminar implementation: Initial Pass.

james7132:
This is an initial crack at #290. This ports over the implementation from `backroll_transport(_udp)` as the `laminar::net::aio` module.

 - Includes the `BidirectionalAsyncChannel<T>`, `Peer`, `Peers<T>` from backroll_transport.
 - Has a `UdpManager` from `backroll_transport_udp` that has been rewritten to have a similar interface to `net::Socket`
 - Added an additional conversion function `with_link_conditioner` to `Peer` that converts the `Peer` and adds tasks that simulate packet loss and latency. The return value is still a `Peer` so the consumer doesn't know of  the conditioner.
 - Added a function to `LinkConditioner` to randomly generate additional latency.
 - Includes a `forward` utility function for doing simple 1:1 channel message forwarding.
 - This design completely removes the need for a `FakeSocket` at this point. A pair of peers can be created in-memory, and `with_link_conditioner` can be used on both sides to achieve the same result.

Potential changes before merging:
 
 - The UDP implementation is included because the base synchronous Laminar uses it, but we may want to put it behind a default feature flag in the future
 - The implementation is closely tied with `bevy_tasks` TaskPool as it's futures executor. This, ideally, should be managed by a feature flag and be runtime agnostic.
 - Unit test the link conditioned peer.
 - More documentation.

Current unknowns:

 - Should the synchronous interface be retained? This current implementation does not break backwards compatibility.
 - If the synchronous interface should be retained,  should the underlying implementation be replaced with the async one?
 - Should `Peer` or one of it's wrappers send one of the `Event` enums? The disconnect notification is already baked into the channel implementation, so `SocketEvent::Disconnect` is not going to be needed; however, connection interruption timeouts should definitely be forwarded.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,9 +20,14 @@ autobenches = false
 edition = "2018"
 
 [dependencies]
+async-channel = "1.6"
+async-net = "1.6.0"
+bevy_tasks = "0.5"
+tracing = "0.1"
 byteorder = "1.4.3"
 crc = "1.8"
 crossbeam-channel = "0.5"
+dashmap = "4.0"
 lazy_static = "1.4"
 log = "0.4"
 rand = "0.8"
@@ -38,6 +43,9 @@ serde = "1.0"
 serde_derive = "1.0"
 quickcheck = "1.0"
 quickcheck_macros = "1.0"
+static_assertions = "1.1"
+futures = "0.3"
+serial_test = "0.5"
 
 [features]
 tester = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ edition = "2018"
 [dependencies]
 async-channel = "1.6"
 async-net = "1.6.0"
+futures-timer = "3.0"
 bevy_tasks = "0.5"
 tracing = "0.1"
 byteorder = "1.4.3"
--- a/src/net.rs
+++ b/src/net.rs
@@ -18,4 +18,5 @@ mod quality;
 mod socket;
 mod virtual_connection;
 
+pub mod aio;
 pub mod constants;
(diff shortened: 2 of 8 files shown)


smokku:
> * The implementation is closely tied with `bevy_tasks` TaskPool as it's futures executor. This, ideally, should be managed by a feature flag and be runtime agnostic.

Second that.

I am using `laminar` purely as a transport protocol (TCP replacement) over [`webrtc-unreliable` ](https://github.com/kyren/webrtc-unreliable). I am handling packet pooling in my own runtime. Pulling in a task runtime would just hurt my compilation time for no gain.

james7132:
> I am using laminar purely as a transport protocol (TCP replacement) over webrtc-unreliable . I am handling packet pooling in my own runtime. Pulling in a task runtime would just hurt my compilation time for no gain.

One potential way to do this is to split out the currently private components (i.e. the packet header parsers) for implementing packet handling into a separate crate (i.e. `laminar-proto`), and then have two separate crates that have a sync/async interface for the actual protocol itself?

TimonPost:
Looks great! Thanks. 

- Do we need futures in its whole or can we go for something like future-core? 
- Why did you chose for benvy-tasks, I see it is a less-dependency library but it is quite new. Is this a stable library yet, are there better alternatives?

james7132:
Futures is only needed for testing. Only using it for futures::block_on right now, so I think the dependency surface area could be reduced.

bevy-tasks was primarily chosen because it supports a multitude of game platforms, including wasm out of the box. However, the only thing needed is a way to launch futures, and it may be useful to use [async-executors](https://docs.rs/async_executors/0.4.1/async_executors/) with feature flags in the future instead.

PR #296, opened 2021-08-09
Add `Packet::into_payload`

Quant1um:
This PR adds a single method to the `Packet` that returns the payload as a `Box<[u8]>` which may be useful to some.


TimonPost (approved):
Look okay!

